PR #9, opened 2023-12-07
Multiple changes to interior order of operations

jarcev:
Likelihoods of individual columns (including the empty one) are now stored as p(c) * ||\nu|| rather than p(c) themselves.

Log-factorial is approximated using Stirling's formula. The constant term is removed, so all likelihoods are increased by log(sqrt(2*\pi)).

Value ||\nu|| * p(c0) is calculated without using matrix multiplication.

The order of additions for ||\nu|| * p(c0) is changed in a way which enables cancellation with the exterior -||\nu||.

The only remaining intermediate value stored in Felsenstein's recursion for p(c0) is now equal to old \tilde{f}-1.

Numerical values in the test are adjusted to be more precise.

For detailed explanations see the pdf doc (to be written soon).


jarcev:
The "pdf doc" in question. https://github.com/acg-team/rust-phylo/files/13629444/Streamline.Notes.pdf

junniest (approved):
Awesome work!

PR #15, opened 2024-07-24
Evo model overhaul

junniest:
Huge evolutionary model refactor.

Goals: 
1. Make whatever is possible generic to reduce duplicate code;
2. Make sure as many constraints as possible are checked by the compiler (e.g. asking for a DNA model with a protein model name is not allowed at compile time due to type mismatches).

Work to do still:
1. PhyoInfo needs refactoring + constraints on models;
2. Generalisation of likelihood costs to be able to use any likelihood for optimisation.


phylo/src/evolutionary_models/mod.rs
```diff
@@ -41,6 +43,22 @@ impl Display for DNAModelType {
...
+impl DNAModelType {
+    pub fn get_model_type(model_name: &str) -> Self {
+        match model_name.to_uppercase().as_str() {
+            "JC69" => DNAModelType::JC69,
+            "K80" => DNAModelType::K80,
+            "HKY" => DNAModelType::HKY,
+            "TN93" => DNAModelType::TN93,
+            "GTR" => DNAModelType::GTR,
+            _ => {
+                warn!("Unknown DNA model requested, defaulting to GTR.");
```
yarcat, line 55: you probably want to log the type as it was passed, otherwise it might be hard to debug. e.g.
```rust
warn!("Unknown DNA model {model_name} requested. Defaulting to GTR.");
```

yarcat: nit: taking into account how easy it is to return `Result<Self>` and handle it, I would also recommend it instead of using defaults -- let the caller handle it

phylo/src/evolutionary_models/mod.rs
```diff
@@ -61,34 +79,55 @@ impl Display for ProteinModelType {
     }
 }
 
-pub trait EvolutionaryModelParameters<T> {
-    fn new(model_type: &T, params: &[f64]) -> Result<Self>
+impl ProteinModelType {
+    pub fn get_model_type(model_name: &str) -> Self {
```
yarcat, line 83: nit: you either wanna name this `from()`, or better use `std::convert` https://doc.rust-lang.org/std/convert/trait.From.html

phylo/src/evolutionary_models/mod.rs
```diff
@@ -41,6 +43,22 @@ impl Display for DNAModelType {
     }
 }
 
+impl DNAModelType {
+    pub fn get_model_type(model_name: &str) -> Self {
```
yarcat, line 47: nit: Please consider naming it `from`, or better use https://doc.rust-lang.org/std/convert/trait.From.html

phylo/src/evolutionary_models/mod.rs
```diff
@@ -61,34 +79,55 @@ impl Display for ProteinModelType {
...
+
+pub trait EvoModelParams {
+    type ModelType;
+    type Parameter;
+    fn new(model: &Self::ModelType, params: &[f64]) -> Result<Self>
     where
         Self: Sized;
-    fn get_value(&self, param_name: &Parameter) -> f64;
-    fn set_value(&mut self, param_name: &Parameter, value: f64);
+    fn parameter_definition(
```
yarcat, line 102: It is absolutely not clear what is this, and what's the `static str` in the resulting vector. I think this is not correctly modeled, but I don't really know the logic.

Having static string smells, like having an enum, which is more type safe. 

If you want to have something printable, it absolutely make sense to add an enum (as suggested above) and having a formatting function, that translates that enum to a string for output.

phylo/src/evolutionary_models/mod.rs
```diff
@@ -61,34 +79,55 @@ impl Display for ProteinModelType {
...
+        model_type: &Self::ModelType,
+    ) -> Vec<(&'static str, Vec<Self::Parameter>)>;
+    fn get_value(&self, param_name: &Self::Parameter) -> f64;
```
yarcat, line 105: I believe in Rust getters don't have `get_` prefix, so in this case it should be just `value`
https://users.rust-lang.org/t/best-practices-on-setters-and-accessor-methods-in-general/66530/4

yarcat: Since you have parameters and frequences, it absolutely would make sense to name these getters smth like `param` (or `param_value`) and `pi` (or probably `freq`)

phylo/src/evolutionary_models/mod.rs
```diff
@@ -61,34 +79,55 @@ impl Display for ProteinModelType {
...
-pub trait EvolutionaryModel<const N: usize> {
-    fn new(model_type: ModelType, params: &[f64]) -> Result<Self>
+pub trait EvolutionaryModel {
+    type ModelType;
+    type Params;
+    fn new(model: Self::ModelType, params: &[f64]) -> Result<Self>
     where
-        Self: std::marker::Sized;
+        Self: Sized;
     fn get_p(&self, time: f64) -> SubstMatrix;
```
yarcat, line 118: ditto: getters should be nouns, not verbs

phylo/src/likelihood/likelihood_tests.rs
```diff
@@ -44,18 +43,19 @@ fn setup_simple_phylo_info(blen_i: f64, blen_j: f64) -> PhyloInfo {
 #[test]
 fn dna_simple_likelihood() {
     let info = setup_simple_phylo_info(1.0, 1.0);
-    let model = DNASubstModel::new(DNA(JC69), &[]).unwrap();
-    let likelihood = DNALikelihoodCost { info: &info };
+    let model = DNASubstModel::new(JC69, &[]).unwrap();
```
yarcat, line 46: optional: Since you might create your model with various options (including an empty smth), it feels like using builders might be a good option. So it would look like
```rust
let model = DNASubstModelBuilder::new(JC69)::build();
```

phylo/src/likelihood/likelihood_tests.rs
```diff
@@ -294,19 +296,19 @@ fn setup_simple_reversibility() -> Vec<PhyloInfo> {
...
+#[case::k80(K80, &[])]
+#[case::hky(HKY, &[0.22, 0.26, 0.33, 0.19, 0.5])]
+#[case::tn93(TN93, &[0.22, 0.26, 0.33, 0.19, 0.5970915, 0.2940435, 0.00135])]
+#[case::gtr(GTR, &[0.1, 0.3, 0.4, 0.2, 5.0, 1.0, 1.0, 1.0, 1.0, 5.0])]
+fn simple_dna_likelihood_reversibility(#[case] model_type: DNAModelType, #[case] params: &[f64]) {
     let info = setup_simple_reversibility();
     let model = DNASubstModel::new(model_type, params).unwrap();
-    let likelihood = DNALikelihoodCost { info: &info[0] };
-    let likelihood_rerooted = DNALikelihoodCost { info: &info[1] };
+    let likelihood = SubstitutionLikelihoodCost::new(&info[0], &model);
```
yarcat, line 307: nit: consider using either named constants, fields, or functions, instead of accessing indices directly, as indices are meaningless to readers (at least to me)

yarcat: some examples:
```rust
let likelihood = SubstitutionLikelihoodCost::new(meaningful_aaa_from(info), &model);
let likelihood_rerooted = SubsctitutionLikelihoodCost::new(meaningfull_bbb_from(info), &model);
```

phylo/src/likelihood/mod.rs
```diff
@@ -1,9 +1,9 @@
 use crate::substitution_models::FreqVector;
 
-pub trait LikelihoodCostFunction<'a, const N: usize> {
+pub trait LikelihoodCostFunction<'a> {
     type Model;
     type Info;
-    fn compute_log_likelihood(&self, model: &Self::Model) -> f64;
+    fn compute_log_likelihood(&self) -> f64;
```
yarcat, line 6: nit: In Rust verbs as methods usually mean that this method modifies the state of this object. Since it doesn't do it, it is better to use nouns
https://rust-lang.github.io/api-guidelines/naming.html

This case is a bit funny, because `log_` wouldn't be a good prefix for the method name (-; But instead of returning f64, you would return a `LogarithmicValue`, for instance and suddenly it might make sense to write it as
```rust
fn likelihood(&self) -> LogarithmicValue;
```
or smth like that

Please note that this is a super-duper nitpicking happening here.

phylo/src/likelihood/mod.rs
```diff
@@ -1,9 +1,9 @@
...
     fn get_empirical_frequencies(&self) -> FreqVector;
```
yarcat, line 7: nit: similarly, `empirical_freq` would be enough

phylo/src/optimisers/branch_length_optimiser.rs
```diff
@@ -0,0 +1,29 @@
+use argmin::core::{CostFunction, Executor, IterState, State};
+use argmin::solver::brent::BrentOpt;
+use log::{debug, info, warn};
+
+use crate::substitution_models::{SubstitutionLikelihoodCost, SubstitutionModel};
+use crate::tree::NodeIdx;
+use crate::Result;
+
+pub(crate) struct SingleBranchOptimiser<'a> {
+    pub(crate) likelihood_cost: &'a SubstitutionLikelihoodCost<'a, 4>,
```
yarcat, line 10: nit: Since you use constants, explaining them with comments/doc-strings makes sense

phylo/src/optimisers/branch_length_optimiser.rs
```diff
@@ -0,0 +1,29 @@
...
+    type Output = f64;
+
+    fn cost(&self, value: &Self::Param) -> Result<Self::Output> {
+        let mut upd_tree = self.likelihood_cost.info.tree.clone();
+        upd_tree.set_branch_length(*self.branch_length, *value);
+        // self.likelihood_cost.info.tree = upd_tree;
+        Ok(-self.likelihood_cost.compute_log_likelihood(self.model).0)
+    }
+
+    fn parallelize(&self) -> bool {
```
yarcat, line 26: nit: `is_*` might be a better name here, as it's a common way to name `bool` getters

phylo/src/optimisers/dna_model_optimiser.rs
```diff
@@ -58,48 +60,47 @@ impl<'a> DNAModelOptimiser<'a> {
 
     pub fn optimise_parameters(
         &self,
-        start_values: &DNASubstParams,
         optimise_freqs: FrequencyOptimisation,
     ) -> Result<(u32, DNASubstParams, f64)> {
```
yarcat, line 64: nit: consider using structs instead of tuples, as it isn't clear what is it returning

phylo/src/optimisers/dna_model_optimiser.rs
```diff
@@ -58,48 +60,47 @@ impl<'a> DNAModelOptimiser<'a> {
...
+        let start_values = self.likelihood_cost.model.params.clone();
```
yarcat, line 65: nit: s/start_values/start_params/

phylo/src/optimisers/dna_model_optimiser.rs
```diff
@@ -58,48 +60,47 @@ impl<'a> DNAModelOptimiser<'a> {
...
         start_values: &DNASubstParams,
-        param_sets: Vec<(&str, Vec<Parameter>)>,
+        param_sets: Vec<(&str, Vec<DNAParameter>)>,
     ) -> Result<(u32, DNASubstParams, f64)> {
         let mut prev_logl = f64::NEG_INFINITY;
-        let mut opt_logl = self
-            .likelihood_cost
-            .compute_log_likelihood(&make_dna_model(start_values.clone()))
-            .0;
+        let mut opt_logl = self.likelihood_cost.compute_log_likelihood().0;
```
yarcat, line 91: nit: tautology in the naming suggests that either self.likelihood_cost should be renamed, or `compute_log_likelihood` method name should be renamed e.g.
```rust
self.likelihood_cost.logarithmic()
```
feels good

phylo/src/optimisers/dna_model_optimiser.rs
```diff
@@ -58,48 +60,47 @@ impl<'a> DNAModelOptimiser<'a> {
...
         info!("Initial logl: {}.", opt_logl);
         let mut opt_params = start_values.clone();
+        let mut model = DNASubstModel::create(&opt_params);
```
yarcat, line 94: thinking out loud: `::create` in Rust conventionally is used to denote that there is an external resource created. Is it the case here? or should `::create` be renamed to `::new` (or `::new_with_params`)?

phylo/src/substitution_models/dna_models/dna_substitution_parameters.rs
```diff
@@ -83,25 +89,19 @@ impl EvolutionaryModelParameters<DNAModelType> for DNASubstParams {
         }
     }
 
+    fn get_pi(&self) -> &FreqVector {
```
yarcat, line 92: btw, in Rust methods and fields can easily have the same name. For instance, this is a fully valid Rust code:
```rust
struct Hello {
    pub world: i32,
}

impl Hello {
    fn world(&self) -> i32 { self.world }
}

fn main() {
    let h = Hello{world: 123};
    println!("{}", h.world());
}
```

phylo/src/substitution_models/dna_models/dna_substitution_parameters.rs
```diff
@@ -83,25 +89,19 @@ impl EvolutionaryModelParameters<DNAModelType> for DNASubstParams {
...
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.print_as(self.model_type))
-    }
-}
 
-impl DNASubstParams {
-    pub(crate) fn parameter_definition(
-        model_type: DNAModelType,
-    ) -> Vec<(&'static str, Vec<Parameter>)> {
+    fn parameter_definition(model_type: &DNAModelType) -> Vec<(&'static str, Vec<DNAParameter>)> {
```
yarcat, line 104: I see now... Yeah, consider returning a vector of Parameters, where parameters should have a method (or a field) that would contain that static str. It is just gonna be more readable

phylo/src/substitution_models/mod.rs
```diff
@@ -104,39 +80,140 @@ where
...
+    fn get_empirical_frequencies(&self) -> FreqVector {
+        let all_counts = self.info.get_counts();
+        let mut total = all_counts.values().sum::<f64>();
+        let index = SubstitutionModel::index(self.model);
+        let mut freqs = FreqVector::from_column_slice(&vec![0.0; Self::Model::N]);
+        for (&char, &count) in all_counts.iter() {
+            freqs += &Self::Model::char_sets()[char as usize].scale(count);
+        }
+        for &char in Self::Model::ALPHABET {
+            if freqs[index[char as usize]] == 0.0 {
```
yarcat, line 203: eq comparison to 0.0 is dangerous. just reminding! you probably want to put a comment in case you really compare to zero values. it might even make sense to compare to f64::Default() just to be explicit

phylo/src/substitution_models/mod.rs
```diff
@@ -104,39 +80,140 @@ where
...
+                freqs[index[char as usize]] += 1.0;
+                total += 1.0;
+            }
+        }
+        freqs.map(|x| x / total)
```
yarcat, line 208: curious if it produces a copy or does in-place mapping for freq?

phylo/src/substitution_models/mod.rs
```diff
@@ -231,8 +315,23 @@ pub struct SubstitutionModelInfo<const N: usize> {
...
+
+    fn new(info: &PhyloInfo, model: &SubstModel) -> Result<Self>
+    where
+        Self: Sized,
+    {
+        Self::new(info, model)
+    }
+
+    fn reset(&mut self) {
+        self.reset();
```
yarcat, line 329: this feels like a recursion. if not, it needs a comment. otherwise it should be fixed.

phylo/src/substitution_models/protein_models/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+
+impl Display for ProteinSubstParams {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.pi)
+    }
+}
+
+impl From<ProteinSubstParams> for Vec<f64> {
+    fn from(val: ProteinSubstParams) -> Self {
+        val.pi.as_slice().to_vec()
```
yarcat, line 76: super-nit: this feels strange. probably just needs a comment. I don't know from this code what's the definition of FreqVector, it feels like it's a named vec[f64]. so, basically, it feels like this code should just return a clone? or maybe even just drop `FreqVector` and leave underlying type? again, I don't know what excactly, but it feels strange, and probably needs a comment

yarcat (approved):

phylo/src/evolutionary_models/mod.rs
junniest: Done.

phylo/src/evolutionary_models/mod.rs
junniest: Done.

phylo/src/evolutionary_models/mod.rs
junniest: Done. While I was at it, I removed all "get" from methods — there were a lot. I also renamed pi() to freqs() to avoid confusing it with p() (and there is also already a q()).

phylo/src/evolutionary_models/mod.rs
junniest: Done.

phylo/src/likelihood/likelihood_tests.rs
junniest: Done.

phylo/src/likelihood/mod.rs
junniest: Done.

phylo/src/optimisers/branch_length_optimiser.rs
junniest: This is a trait implementation, paralellize is part of the CostFunction trait.

phylo/src/substitution_models/dna_models/dna_substitution_parameters.rs
junniest: Done.

phylo/src/substitution_models/mod.rs
junniest: This wasn't recursing because it was calling an internal implementation, but it was completely unnecessary nonetheless. Removed.

phylo/src/optimisers/dna_model_optimiser.rs
junniest: Done.

phylo/src/substitution_models/protein_models/mod.rs
junniest: This is so strange only because, in this case, all the parameters are defined in a vector. For anything other than the protein substitution parameters, this would have more parameters that are not stored in a single vector. And the FreqVector is a DVector<f64> from nalgebra by type, so not the same as Vec<f64>, confusingly enough. That said, I realised this is absolutely unnecessary and I can remove all implementations of this trait from my parameters, which I have now done (:

phylo/src/likelihood/mod.rs
junniest: > In Rust verbs as methods usually mean that this method modifies the state of this object. Since it doesn't do it, it is better to use nouns

I cannot find this convention anywhere, neither at the provided link nor online. I would love to see it; it makes a lot of sense.

That said, the method name is, in fact, garbage (: It will be refactored in the next PR, so will leave it as is here.

phylo/src/optimisers/dna_model_optimiser.rs
junniest: I agree that it should be renamed, but I don't agree with the proposed rename. Will think about it and rename/refactor in next PR.

phylo/src/optimisers/dna_model_optimiser.rs
junniest: Agreed, will be done in next PR where some of the optimisation is standardised.

phylo/src/optimisers/branch_length_optimiser.rs
junniest: Agreed. This will be generalised away in the next PR, but I will explain constants where they still appear.

phylo/src/likelihood/likelihood_tests.rs
junniest: This is a good idea, but I need to think about how to make it work properly as an API for when the code comes together. This will happen soon, but not in this PR.

phylo/src/optimisers/dna_model_optimiser.rs
junniest: I will rethink these names. Potentially with a better API only one of new/create will be necessary (currently there's both and it's pointless because they do kind of the same thing).

phylo/src/substitution_models/dna_models/dna_substitution_parameters.rs
junniest: The problem here is that this defines parameter sets that should be all set to equal and optimised together. 

This is an example of the output of that function:
```
vec![("alpha1", vec![Rtc]), ("alpha2", vec![Rag]), ("beta", vec![Rta, Rtg, Rca, Rcg])]
```
Otherwise, e.g., the Rta and Rtg parameters will be decoupled, changing the model to a different one in our context. I will think about how to improve this, but for now I need to prioritise the next PR.

phylo/src/evolutionary_models/mod.rs
junniest: I will either refactor this in the next PRs or I will add a docstring. Can't think of how to do this better at the moment (also see comment below).

PR #16, opened 2024-08-05
Branch len optimisation

junniest:
Two birds with one PR (bad call, I know):
Initial branch length optimisation under the PIP;
Refactor of the Tree structure to only use one list of nodes instead of leaves and internals separately.

### DIFF
--- a/phylo/src/optimisers/branch_length_optimiser.rs
+++ b/phylo/src/optimisers/branch_length_optimiser.rs
@@ -5,13 +5,13 @@ use log::{debug, info};
 use crate::phylo_info::PhyloInfo;
 use crate::pip_model::{PIPLikelihoodCost, PIPModel};
 use crate::substitution_models::SubstitutionModel;
-use crate::tree::{Node, NodeIdx, Tree};
+use crate::tree::{NodeIdx, Tree};
 use crate::Result;
 
 pub(crate) struct SingleBranchOptimiser<'a, SubstModel: SubstitutionModel> {
     pub(crate) model: &'a PIPModel<SubstModel>,
     pub(crate) phylo_info: &'a PhyloInfo,
-    pub(crate) branch: NodeIdx,
+    pub(crate) branch: &'a NodeIdx,
 }
 
 impl<'a, SubstModel: SubstitutionModel + Clone> CostFunction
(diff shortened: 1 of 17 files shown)


phylo/src/optimisers/branch_length_optimiser.rs
```diff
@@ -1,29 +1,113 @@
...
+{
+    pub fn new(model: &'a PIPModel<SubstModel>, phylo_info: &PhyloInfo) -> Self {
+        BranchOptimiser {
+            epsilon: 1e-3,
+            model,
+            phylo_info: phylo_info.clone(),
+        }
+    }
+
+    pub fn optimise_parameters(&self) -> Result<(u32, Tree, f64, f64)> {
```
yarcat, line 60: nit: consider using structures instead of tuples. its to hard to understand what are these u32,Tree,f64,f64. Everything except for Tree is too generic.

yarcat: Alternatively, you can put a comment e.g. `// iterations, tree copy, init, opt` or whatever makes sense

phylo/src/optimisers/branch_length_optimiser.rs
```diff
@@ -1,29 +1,113 @@
...
+            for branch in &nodes {
+                let (logl, length) = self.optimise_branch(&lc, branch)?;
+                opt_logl = logl;
+                lc.info.tree.set_branch_length(branch, length);
+                debug!(
+                    "Optimised branch length {} to value {} with logl {}",
+                    branch, length, opt_logl
+                );
+            }
+            iters += 1;
```
yarcat, line 87: nit: I suggest increasing that as the first thing in the loop

phylo/src/optimisers/branch_length_optimiser.rs
```diff
@@ -1,29 +1,113 @@
...
+            opt_logl, iters
+        );
+        Ok((iters, lc.info.tree.clone(), init_logl, opt_logl))
+    }
+
+    fn optimise_branch(
+        &self,
+        lc: &PIPLikelihoodCost<SubstModel>,
+        branch: &NodeIdx,
+    ) -> Result<(f64, f64)> {
```
yarcat, line 100: ditto. probably a comment `// best_cost, best_param` would be good enough

yarcat (approved):

phylo/src/optimisers/branch_length_optimiser.rs
junniest: Good point, done.

phylo/src/optimisers/branch_length_optimiser.rs
junniest: Will be fixed in next PR with an optimisation result struct that all optimisers will use.

phylo/src/optimisers/branch_length_optimiser.rs
junniest: Will also be fixed in next PR, either with a comment or with something more meaningful.

PR #17, opened 2024-08-14
Msa refactor

junniest:
Huge refactor to use Alignment structures throughout for computation.


phylo/src/alignment/alignment_builder.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    pub fn msa(mut self, msa: InternalMapping) -> Self {
+        self.node_map = msa;
+        self
+    }
+
+    pub fn build(self) -> Result<Alignment> {
+        if self.node_map.is_empty() {
```
yarcat, line 31: nit: I would recommend reversing the condition here, as it would reduce the levels in the function:
```rust
if !self.node_map.is_empty() {
  self.build_from_map()
} else if self.seqs.aligned {
  self.build_from_seqs()
} else {
  self.build_from_unaligned()
}
```

phylo/src/alignment/alignment_builder.rs
```diff
@@ -0,0 +1,121 @@
...
+            } else {
+                self.build_from_unaligned()
+            }
+        } else {
+            self.build_from_map()
+        }
+    }
+
+    /// This assumes that the tree structure matches the alignment structure and that the sequences are aligned.
+    fn build_from_seqs(self) -> Result<Alignment> {
```
yarcat, line 43: jfyi: usually I would say that `from_seqs`, `from_unaligned` and `from_map` is enough. not need to use verbs. here it's a builder, so just check whatever is recommended in rust

yarcat: Based on `HasMap::with_capacity` below, I'd say that `from_xxx` is the way to go

phylo/src/alignment/alignment_builder.rs
```diff
@@ -0,0 +1,121 @@
...
+        };
+        let leaf_map = alignment.compile_leaf_map(self.tree.root, self.tree)?;
+        alignment.leaf_map = leaf_map;
+        alignment.seqs = self.seqs.without_gaps();
+        Ok(alignment)
+    }
+
+    fn stack_maps(msa_len: usize, map_x: &Mapping, map_y: &Mapping) -> Mapping {
+        let mut map = Vec::with_capacity(msa_len);
+        let mut ind: usize = 0;
```
yarcat, line 96: nit: `let mut ind = 0usize`

phylo/src/alignment/alignment_builder.rs
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+        map
+    }
+
+    fn clear_common_gaps(msa_len: usize, map_x: &Mapping, map_y: &Mapping) -> PairwiseAlignment {
+        let mut upd_map_x = Vec::with_capacity(msa_len);
+        let mut upd_map_y = Vec::with_capacity(msa_len);
+        for (x, y) in map_x.iter().zip(map_y.iter()) {
+            if x.is_none() && y.is_none() {
```
yarcat, line 112: nit: you could make it shorter by reverting the condition:
```rust
if x.is_some() || y.is_some() {
  upd_map_x.push(*x);
  upd_map_y.push(*y);
}
```

yarcat: alternatively, you can remove the `else` clause, as you early continue

phylo/src/alignment/sequences.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    pub fn get(&self, idx: usize) -> &Record {
+        &self.s[idx]
+    }
+
+    pub fn get_mut(&mut self, idx: usize) -> &mut Record {
+        &mut self.s[idx]
+    }
+
+    pub fn get_by_id(&self, id: &str) -> &Record {
```
yarcat, line 56: nit: this would be `from_id`

phylo/src/alignment/alignment_builder.rs
junniest: Yes, makes perfect sense (;

phylo/src/alignment/alignment_builder.rs
junniest: Done.

phylo/src/alignment/alignment_builder.rs
junniest: lol of course, done.

phylo/src/alignment/sequences.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    pub fn len(&self) -> usize {
+        self.s.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.s.is_empty()
+    }
+
+    pub fn get(&self, idx: usize) -> &Record {
```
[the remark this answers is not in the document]
junniest: At least according to ChatGPT, Rust actually prefers getters to be called by the name of the field itself, so I should rather use record() or record_mut(). When I look through the codebase it doesn't seem like there's a lot of ref() and ref_mut() going on. In this case, I prefer
` pub fn record(&self, idx: usize) -> &Record {}`
What do you think?

phylo/src/alignment/sequences.rs
junniest: This method is a getter, but `from_id()` by convention would be a constructor and would not take self. It is now renamed to `record_by_id()` as shown above.

phylo/src/alignment/sequences.rs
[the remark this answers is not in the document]
junniest: changed to add a non-panicking version of `record_by_id` just in case:

```
    pub fn record(&self, idx: usize) -> &Record {
        &self.s[idx]
    }

    pub fn record_mut(&mut self, idx: usize) -> &mut Record {
        &mut self.s[idx]
    }

    pub fn record_by_id(&self, id: &str) -> &Record {
        self.s
            .iter()
            .find(|r| r.id() == id)
            .unwrap_or_else(|| panic!("Sequence with id {} not found", id))
    }

    pub fn try_record_by_id(&self, id: &str) -> Result<&Record> {
        let rec = self.s.iter().find(|r| r.id() == id);
        match rec {
            Some(r) => Ok(r),
            None => bail!("Sequence with id {} not found", id),
        }
    }
```

phylo/src/alignment/alignment_builder.rs
junniest: No, I think this is worse. ChatGPT suggests using action-oriented and context-specific names, so this is the current version:
```
    pub fn build(self) -> Result<Alignment> {
        if !self.node_map.is_empty() {
            self.compile_from_map()
        } else if self.seqs.aligned {
            self.reconstruct_from_aligned_seqs()
        } else {
            self.align_unaligned_seqs()
        }
    }
```

I prefer this.

PR #21, opened 2025-01-07
Model refactor

junniest:
Refactor to make sure that the PhyloCostFunction doesn't have a random "reset" parameter.
Instead, the function should handle any change in model parameters or the tree internally (when update_tree, set_param or set_freqs are called).


phylo/src/optimisers/model_optimiser.rs
```diff
@@ -1,132 +1,119 @@
...
+            c: RefCell::new(cost),
             freq_opt,
         }
     }
 
-    fn run(self) -> Result<EvoModelOptimisationResult<EM>> {
-        let mut model = self.model.clone();
-        let initial_logl = model.cost(&self.info, true);
-        info!("Optimising {} parameters.", model.description());
+    pub fn run(mut self) -> Result<EvoModelOptimisationResult<C>> {
```
yarcat, line 28: Please double check `mut`. It doesn't feel right.

phylo/src/substitution_models/mod.rs
```diff
@@ -115,146 +166,93 @@ where
...
+where
+    SubstModel<Q>: EvoModel,
+{
+    fn cost(&self) -> f64 {
+        self.logl(&self.info)
     }
-
-    fn p(&self, time: f64) -> SubstMatrix {
-        SubstitutionModel::p(self, time)
+    fn update_tree(&mut self, tree: &Tree, dirty_nodes: &[NodeIdx]) {
```
yarcat, line 207: Why are you passing reference and then immediately cloning the value? Consider changing the interface so that it accepts a value, and cloning it should be a business of the caller. It would be cleaner that way because otherwise your `tree` is virtually unused (it's clone is used), making the function interface wrong.

If this is not the case, then the comments are missing somewhere.

yarcat (approved):

phylo/src/likelihood/mod.rs
```diff
@@ -1,8 +1,15 @@
-use crate::phylo_info::PhyloInfo;
+use crate::substitution_models::FreqVector;
+use crate::tree::{NodeIdx, Tree};
 
 pub trait PhyloCostFunction {
-    fn cost(&self, info: &PhyloInfo, reset: bool) -> f64;
-    fn reset(&self);
+    fn cost(&self) -> f64;
+    fn update_tree(&mut self, tree: &Tree, dirty_nodes: &[NodeIdx]);
```
yarcat, line 6: why is this one `update_tree`, but then below there's `set_param`? while the code actually shows that the tree is set rather than updated

yarcat: here I'm trying to understand the logic behind `update` and `set`

phylo/src/optimisers/model_optimiser.rs
junniest: Removed, it was not necessary. I got lost in RefCells (:

phylo/src/likelihood/mod.rs
junniest: That was meant to show that the tree cannot change arbitrarily: the tree is bound to the alignment internally, so it can not be set to a random tree; it has to be the old tree with some changes, indicated by the dirty nodes. But I see the confusion, just not sure how to resolve that.

phylo/src/substitution_models/mod.rs
junniest: Done.

phylo/src/likelihood/mod.rs
junniest: This will get reworked in the next iteration, added a comment to clarify that the tips of the tree must be the same, but this is not enforced in the code yet.

phylo/src/likelihood/mod.rs
```diff
@@ -1,8 +1,15 @@
-use crate::phylo_info::PhyloInfo;
+use crate::substitution_models::FreqVector;
+use crate::tree::{NodeIdx, Tree};
 
 pub trait PhyloCostFunction {
```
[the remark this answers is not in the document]
junniest: Split the trait in 2: ModelSearchCost and TreeSearchCost.

PR #27, opened 2025-02-01
wrote test for without_gaps, removed duplication

### LINKED ISSUE #26: Rework the aligned and unaligned seqs in tests of alignment
The `unaligned_seqs()` in `alignment/tests.rc` is duplicate code with `Sequences::without_gaps`. Since the changes in #25 will include sequences for ancestors and I also want to check for unaligned sequences, I want to reuse the `Sequences::without_gaps` and not introduce more duplicate code.

Tasks:
- use `Sequences::without_gaps` in the tests
- write tests for `Sequences::without_gaps`

MattesMrzik:
Closes #26 
- Wrote the test for `Sequences::without_gaps(&self)`
- removed duplication in `alignment/tests.rc`
  - as a consequence the random shuffling of the unaligned sequences got removed, should this be reimplemented?

### DIFF
--- a/phylo/src/alignment/tests.rs
+++ b/phylo/src/alignment/tests.rs
@@ -42,26 +42,6 @@ fn test_tree() -> Tree {
 }
 
 #[cfg(test)]
-fn unaligned_seqs() -> Sequences {
-    let unaligned_seqs = aligned_seqs(&["A0", "B1", "C2", "D3", "E4"])
-        .s
-        .into_iter()
-        .map(|rec| {
-            record!(
-                rec.id(),
-                rec.desc(),
-                &rec.seq()
-                    .iter()
-                    .filter(|&c| c != &b'-')
-                    .copied()
-                    .collect::<Vec<u8>>()
-            )
-        })
-        .choose_multiple(&mut thread_rng(), 5);
-    Sequences::new(unaligned_seqs)
-}
-
-#[cfg(test)]
 fn maps() -> (InternalMapping, LeafMapping) {
     let aligned_seqs = aligned_seqs(&["A0", "B1", "C2", "D3", "E4"]);
     (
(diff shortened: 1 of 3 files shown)


phylo/src/alignment/tests.rs
```diff
@@ -144,10 +124,41 @@ fn sequences_with_alphabet() {
...
+        record!("E4", Some("E4 sequence"), b"-A-AAA"),
+    ];
+    let seqs_without_gaps = vec![
+        record!("A0", Some("A0 sequence"), b"AAAAAA"),
+        record!("B1", Some("B1 sequence"), b"AA"),
+        record!("C2", Some("C2 sequence"), b"AAA"),
+        record!("D3", Some("D3 sequence"), b"AA"),
+        record!("E4", Some("E4 sequence"), b"AAAA"),
+    ];
+    let sequences = Sequences::new(seqs.clone());
```
junniest, line 144: 2 points:
1. You don't need to clone here, the `seqs` variable is never used again.
2. Because `seqs` is never used again, it probably makes more sense to not create it altogether, making line 144:
`let sequences = Sequences::new(vec![record!("A0", Some("A0 sequence"), b"AAAAAA"), ...]);`

phylo/src/alignment/tests.rs
```diff
@@ -144,10 +124,41 @@ fn sequences_with_alphabet() {
...
+fn sequences_without_gaps() {
+    // arrange
+    let seqs = vec![
+        record!("A0", Some("A0 sequence"), b"AAAAAA"),
+        record!("B1", Some("B1 sequence"), b"---A-A"),
+        record!("C2", Some("C2 sequence"), b"AA---A"),
+        record!("D3", Some("D3 sequence"), b"---A-A"),
+        record!("E4", Some("E4 sequence"), b"-A-AAA"),
+    ];
+    let seqs_without_gaps = vec![
```
junniest, line 137: I think this can be called `expected_seqs` to show what is the data you're using as test input and what is not data, but the expected output.

phylo/src/alignment/tests.rs
```diff
@@ -144,10 +124,41 @@ fn sequences_with_alphabet() {
...
+        record!("A0", Some("A0 sequence"), b"AAAAAA"),
+        record!("B1", Some("B1 sequence"), b"AA"),
+        record!("C2", Some("C2 sequence"), b"AAA"),
+        record!("D3", Some("D3 sequence"), b"AA"),
+        record!("E4", Some("E4 sequence"), b"AAAA"),
+    ];
+    let sequences = Sequences::new(seqs.clone());
+
+    // act
+    let sequences = sequences.without_gaps();
```
junniest, line 147: Same thought here as with the previous comment, the naming here is confusing. We can have this in two lines, but then I'd rename `sequences` to `msa`:
`let msa = Sequences::new(vec![record!("A0", Some("A0 sequence"), b"AAAAAA"), ...]);
let sequences = msa.without_gaps();`
Or, squish all of this to remove intermediate variables altogether:
`let sequences = Sequences::new(vec![record!("A0", Some("A0 sequence"), b"AAAAAA"), ...]).without_gaps();`

junniest (commented):
Thanks for noticing the duplication and removing it! That's great.
I think that my `without_gaps()` method is far from ideal also in terms of naming, so if you have an idea on how to make it better, I'd be happy to hear it.

I do unfortunately think that it's best to add back the random shuffle of sequences because it was there to make sure that the MSA compilation doesn't break when the order of sequences doesn't match the tree. We can do that by adding some form of `.choose_multiple(&mut thread_rng(), 5);` to the respective locations where unaligned sequences appear (lines 161, 177, 189, 227, 242, 260).

phylo/src/alignment/tests.rs
MattesMrzik: Thanks. That's better.

phylo/src/alignment/tests.rs
MattesMrzik: Yes, I agree. But I wouldn't call it `msa` but `aligned_seqs`, because it is not a `Alignment` struct. What do you think of that?

MattesMrzik:
> I do unfortunately think that it's best to add back the random shuffle of sequences because it was there to make sure that the MSA compilation doesn't break when the order of sequences doesn't match the tree. We can do that by adding some form of `.choose_multiple(&mut thread_rng(), 5);` to the respective locations where unaligned sequences appear (lines 161, 177, 189, 227, 242, 260).

Okay, makes sense. Then something like this?
```rust 
let mut unaligned_seqs = aligned_seqs(&["A0", "B1", "C2", "D3", "E4"]).without_gaps();
unaligned_seqs.s = unaligned_seqs.s.choose_multiple(&mut thread_rng(), 5).cloned().collect();
```

Or implement a `#[test(cfg)]` method that shuffles the seqs vector of the `Sequences` instance. I would prefer this to the code snippet above.

(I would also rename the method `without_gaps` to `remove_gaps`)

phylo/src/alignment/tests.rs
junniest: Agree! You're absolutely right, it's not an `msa`, it's `aligned_seqs`.

junniest (approved):
LGTM.

PR #35, opened 2025-02-14
Fixes for binary

junniest:
Fixing EvoModel and QMatrix traits to be dyn compatible so that the binary can use this crate as intended.


junniest:
Unfortunately this ended up including a bunch of commits I forgot about ): [821db44] to [e8a62d8] are from before and are just minor cleanup.

phylo/src/optimisers/model_optimiser.rs
```diff
@@ -38,14 +38,20 @@ impl<C: ModelSearchCost + Display + Clone> ModelOptimiser<C> {
...
             iterations += 1;
             debug!("Iteration: {}", iterations);
+            let parameters = self.c.borrow().params().to_vec();
             prev_logl = final_logl;
-            for (param, value) in parameters.iter().enumerate() {
-                let (value, logl) = self.opt_parameter(param, *value)?;
+            for (param, start_value) in parameters.iter().enumerate() {
+                debug!(
+                    "Optimising parameter {:?} from value {} with logl {}",
+                    param, start_value, final_logl
```
MattesMrzik, line 49: What about renaming `final_logl` to sth like `current_best`? Since it's being iteratively improved.

MattesMrzik (dismissed):
Seems good. Did not spot major issues.

phylo/src/likelihood/tests.rs
```diff
@@ -136,3 +134,89 @@ fn protein_pip_search_costs_equal() {
...
+fn alphabet_mismatch_subst_model_template<Q: QMatrix + QMatrixMaker>(
+    alpha: Alphabet,
+    freqs: &[f64],
+    params: &[f64],
+) {
+    // https://molevolworkshop.github.io/faculty/huelsenbeck/pdf/WoodsHoleHandout.pdf
+    let fldr = Path::new("./data");
+    let records =
+        read_sequences_from_file(&fldr.join("Huelsenbeck_example_long_DNA.fasta")).unwrap();
+    let seqs = Sequences::with_alphabet(records.clone(), alpha);
```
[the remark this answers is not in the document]
junniest: true (: It's not too bad since this is a unit test, but I agree that it's better to remove all unnecessary stuff like that.

phylo/src/pip_model/tests.rs
```diff
@@ -165,25 +153,23 @@ fn pip_protein_subst_rates() {
...
-            pip.q.row(pip.index[char as usize]).sum(),
-            0.0,
-            epsilon = 1e-14
-        );
+        let mut sum = 0.0;
+        for &other_char in chars {
+            sum += pip.rate(char, other_char);
+        }
+        assert_relative_eq!(sum, 0.0, epsilon = 1e-14);
         assert_relative_eq!(pip.rate(char, b'-'), params[1]);
```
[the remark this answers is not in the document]
junniest: Good point, replaced all uses with the const now.

phylo/src/optimisers/model_optimiser.rs
junniest: Good point. And it's not necessarily the logl, it's just a cost function. Renamed the variables accordingly and changed the log messages to match.

phylo/src/optimisers/mod.rs
```diff
@@ -7,16 +7,21 @@ pub use model_optimiser::*;
 pub mod topo_optimiser;
 pub use topo_optimiser::*;
 
+pub struct SingleValOptResult {
+    pub final_cost: f64,
+    pub value: f64,
```
MattesMrzik, line 12: Just locking at this struct the `value` seems non-descriptive (maybe in contexts it makes sense, but a comment on what is usually stored there would be nice)

MattesMrzik (commented):
Looks good to me

MattesMrzik (approved):

PR #42, opened 2025-03-28
Fix/pip likelihood crash

junniest:
In extreme cases some of the branch lengths tended to infinity which made the probability of a single site underflow and become zero. Rough fix for this by resetting the site probabilities at the root from 0.0 to f64::MIN_POSITIVE to limit the likelihood at a meaningful value and to make sure that making a branch infinitely longer doesn't make likelihood go to -Inf.

### DIFF
--- a/phylo/src/pip_model/tests.rs
+++ b/phylo/src/pip_model/tests.rs
@@ -824,8 +824,6 @@ fn blen_leading_to_minusinf() {
     );
     let tree = tree!("((284811:0.0000000000000002,(284593:0.1,(237561:0.3,(284812:0.3,(284813:400.9,284591:0.2):40000000000000.2):0.05):0.1):0.04):0);");
     let info = PIB::build_from_objects(sequences, tree).unwrap();
-    //0.0012197623652386346
-    //0.0012197623652386337
 
     let model = PIPModel::<WAG>::new(&[], &[]);
     let c = PIPB::new(model, info).build().unwrap();


phylo/src/pip_model/mod.rs
```diff
@@ -317,7 +317,23 @@ impl<Q: QMatrix> PIPCost<Q> {
         let root_idx = usize::from(&self.info.tree.root);
         let msa_length = self.info.msa.len();
 
-        tmp.pnu[root_idx].map(|x| x.ln()).sum() + tmp.c0_pnu[root_idx]
+        // In certain scenarios (e.g. a completely unrelated sequence, see data/p105.msa.fa)
```
junniest, line 320: @MattesMrzik Could you take a look here to make sure it makes sense? (:

MattesMrzik: The checks `x == 0.0 || x.is_subnormal()` seem good to me. Is the `x.is_nan()` something you added just for good measure? If this condition is not produced by known circumstances, then it might be better to not include it, bc now it might cover up some unknown bug that produces `nan`.

phylo/src/pip_model/tests.rs
```diff
@@ -794,3 +794,42 @@ fn logl_not_inf_for_empty_col() {
...
+            record!("237561", b"W"),
+            record!("284591", b"W"),
+            record!("284812", b"W"),
+        ],
+        protein_alphabet(),
+    );
+    let tree = tree!("((284811:0.0000000000000002,(284593:0.1,(237561:0.3,(284812:0.3,(284813:400.9,284591:0.2):40000000000000.2):0.05):0.1):0.04):0);");
+    let info = PIB::build_from_objects(sequences, tree).unwrap();
+    //0.0012197623652386346
+    //0.0012197623652386337
```
MattesMrzik, line 828: What are these comments? Either remove them or add explanation:)

MattesMrzik (changes requested):
Got just two little comments.

phylo/src/pip_model/mod.rs
junniest: Good call. I did add it for good measure, but you are right, NaN probabilities have a very different meaning to this, will remove.

phylo/src/pip_model/tests.rs
junniest: Good point, these were some values I was comparing and forgot to delete, my bad!

MattesMrzik (approved):

PR #48, opened 2025-04-15
Feature/Implements ancestral alignment

### LINKED ISSUE #25: Alignment with ancestral sequences for TKF92
To make the tree search work with TKF92 (while not integrating over all possible ancestral seq) we need to include the ancestral sequences in the alignment. 

- For this, a new struct `AncestralAlignment` should be implemented. 
The `pub type InternalMapping = HashMap<NodeIdx, PairwiseAlignment>;` of the current `Alignment` struct is not really applicable with ancestors. We either change this to an MSA with 3 `Mapping`s or drop it for the `AncestralAlignment` struct.
- Additionally, the `PhyloInfo` needs to be reworked to either hold an `Alignment` or an `AncestralAlignment`. Alternatively, a new struct can be implemented that holds the `AncestralAlignment` instead of a normal `Alignment`. 

(If it turns out, that we can find the maximum likely ancestors for a given leaf alignment in reasonable time, then we might not need this new struct.)

MattesMrzik:
## This PR is DISCONTINUED! It is continued in #53 

- implements AncestralAlignment struct
- can be built from alignment of all nodes
- can be built from alignment of only leafs (does parismony reconstruction of ancestors)
- wrote tests

## TODOs:
- currently I have a type (SeqMapping) that is the same as `LeafMapping` that I need because in TKF the internal nodes will also have sequences. Either we keep this duplication or try to come up with a more general name for the type that can both be used in the `Alignment` and `AncestralAlignment`
- The internal mapping that was used in `Alignment` which may be used for the realignment step is not implemented yet for `AncestralAlignment`.

### DIFF
--- a/phylo/src/alignment/alignment_builder.rs
+++ b/phylo/src/alignment/alignment_builder.rs
@@ -5,7 +5,10 @@ use bio::io::fasta::Record;
 use log::warn;
 
 use crate::align;
-use crate::alignment::{Alignment, InternalMapping, Mapping, PairwiseAlignment, Sequences};
+use crate::alignment::{
+    Alignment, AncestralAlignment, InternalMapping, Mapping, PairwiseAlignment, SeqMapping,
+    Sequences,
+};
 use crate::alphabets::GAP;
 use crate::tree::{NodeIdx, NodeIdx::Internal as Int, NodeIdx::Leaf, Tree};
 use crate::Result;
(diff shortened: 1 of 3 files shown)


phylo/src/alignment/alignment_builder.rs
```diff
@@ -123,3 +126,134 @@ impl<'a> AlignmentBuilder<'a> {
...
+        let n_nodes = self.tree.len();
+
+        // inferring ancestral sequences for every site independently:
+        // insertion location is latest common ancestors of all non-gap characters
+        // deletion location is node n such that every leaf in the subtree rooted in n
+        // is a gap and the parent of n is not a deletion location
+        let mut all_maps = leaf_maps;
+        // counter[node] keeps track of the sequence index during the procedurally generated mapping for the node
+        let mut counter = vec![0; n_nodes];
+        for site in 0..alignment_len {
```
junniest, line 158: **IMPORTANT:** This is an inefficient way to iterate over our data. It is not critical for performance in this case as this will only get executed once (when we read in the input, right?), but this pattern will be critical in other places.
Briefly: you want to always iterate over the tree first and sites second, rather than the other way around, sites first then nodes. this has to do with caching, jumping around different nodes all the time will make caching impossible for large trees. We can discuss this if needed, writing the full story here is too tedious.

phylo/src/alignment/alignment_builder.rs
```diff
@@ -123,3 +126,134 @@ impl<'a> AlignmentBuilder<'a> {
...
+                        !has_char[usize::from(children[0])] && !has_char[usize::from(children[1])];
+
+                    let char_was_chosen_for_parent = match parent {
+                        None => false,
+                        Some(parent) => all_maps[parent][site].is_some(),
+                    };
+                    let must_choose_char =
+                        (both_have_chars || char_was_chosen_for_parent) && !both_are_gap;
+                    // appending to the mapping of the node
+                    match all_maps.get_mut(node) {
```
junniest, line 193: I find this part confusing, but it will change when you address the important comment above, so we can discuss then.

phylo/src/alignment/alignment_builder.rs
```diff
@@ -123,3 +126,134 @@ impl<'a> AlignmentBuilder<'a> {
...
+                self.build_from_only_aligned_leafs()
+            } else {
+                Err(
+                    "The number of sequences does not match the number of nodes nor the number of leafs \
+                    in the tree, which is required for an ancestral alignment."
+                        .to_string(),
+                )
+            }
+        } else {
+            Err("Unaligned sequences are not yet supported.".to_string())
```
junniest, line 256: Again, `bail!("msg")`

phylo/src/alignment/mod.rs
```diff
@@ -174,6 +175,56 @@ impl Alignment {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct AncestralAlignment {
+    pub(crate) seqs: Sequences,
+    seq_map: SeqMapping,
+    // TODO: implement  node_map: InternalMapping,
+    #[allow(dead_code)]
+    pub(crate) leaf_encoding: HashMap<String, DMatrix<f64>>,
```
junniest, line 184: Why does this have `#[allow(dead_code)]`? If it is not used at all, we need to remove it. It should probably be removed from `Alignment` too if that is the case.

phylo/src/alignment/mod.rs
```diff
@@ -174,6 +175,56 @@ impl Alignment {
...
+            .values()
+            .next()
+            .map(|map| map.len())
+            .unwrap_or(0)
+    }
+    pub fn seq_count(&self) -> usize {
+        self.seq_map.len()
+    }
+
+    pub fn update_nodes(&mut self, new_nodes: SeqMapping) {
```
junniest, line 200: Why is this part of the public interface of this structure? If someone else is using this, how would they know what is the exact expected contents of `new_nodes` that would not break this structure? There are too many invalid inputs to this for this method to be public. Either it needs restrictions to validate the input, or, more likely, it should not be `pub`. By default, most methods should not be `pub`, and if they are, there should not be a way to break the structure when using the method.

phylo/src/alignment/alignment_builder.rs
```diff
@@ -123,3 +126,134 @@ impl<'a> AlignmentBuilder<'a> {
...
+    tree: &'a Tree,
+    seqs: Sequences,
+}
+
+impl<'a> AncestralAlignmentBuilder<'a> {
+    pub fn new(tree: &'a Tree, seqs: Sequences) -> AncestralAlignmentBuilder<'a> {
+        AncestralAlignmentBuilder { tree, seqs }
+    }
+
+    fn build_from_only_aligned_leafs(self) -> std::result::Result<AncestralAlignment, String> {
```
junniest, line 140: This method is way too long to be a single method and the name is descriptive but also way too long. I would suggest at least `build_from_aligned_leaves`, but it would probably be better to make a generic `build` be the entry point (as in `Alignment`) and then have it choose which method is appropriate (only leaf nodes given vs all nodes). This function should probably also be called something other than `build` potentially, since what it does is not actually building an alignment, it's just reading it in and transforming it into the data structure you need.

phylo/src/alignment/mod.rs
```diff
@@ -174,6 +175,56 @@ impl Alignment {
...
+        for new_node in new_nodes.keys() {
+            assert!(
```
junniest, line 202: Also, we should not use asserts in production code, they're only there when a true panic should happen (impossible values given to a method). In general we want to avoid panics when a method could have been given incorrect parameters. Plus, this only catches one kind of incorrect values, whereas there are plenty more options to provide wrong input.

phylo/src/alignment/alignment_builder.rs
```diff
@@ -123,3 +126,134 @@ impl<'a> AlignmentBuilder<'a> {
...
+        let seqs = self.seqs.into_gapless();
+        let leaf_encoding = seqs.generate_leaf_encoding();
+        Ok(AncestralAlignment {
+            seqs,
+            seq_map: all_maps,
+            leaf_encoding,
+        })
+    }
+
+    fn build_from_aligned_seqs_with_ancestors(
```
junniest, line 225: again, name too long, rename to e.g. `build_from_aligned_ancestors` or something along those lines.

phylo/src/alignment/alignment_builder.rs
```diff
@@ -123,3 +126,134 @@ impl<'a> AlignmentBuilder<'a> {
...
+        self,
+    ) -> std::result::Result<AncestralAlignment, String> {
```
junniest, line 227: Change `std::result::Result<AncestralAlignment, String>` to `Result<AncestralAlignment>`.

phylo/src/alignment/mod.rs
```diff
@@ -174,6 +175,56 @@ impl Alignment {
...
+                self.seq_map.contains_key(new_node),
+                "The node that is to be updated ({new_node}) is not in the tree.",
+            )
+        }
+        self.seq_map.extend(new_nodes);
+    }
+}
+
+impl Display for AncestralAlignment {
```
junniest, line 211: I don't understand what is the Display format here. I would recommend printing the alignment in some standard format, e.g. fasta, rather than a custom format for sequences. This way this output can be viewed just with an alignment viewer easily.
The `Display` implementation in `Alignment` isn't great, but I am just using the `bio::io::fasta::Record` printing capability which will print out fasta sequences.

phylo/src/alignment/mod.rs
```diff
@@ -174,6 +175,56 @@ impl Alignment {
...
+        let mut internal = String::new();
+        for (x, v) in self.seq_map.iter() {
+            if !matches!(x, Int(_)) {
+                continue;
+            }
+            internal.push_str(&format!("{x}, "));
+            for opt in v {
+                internal.push(if opt.is_some() { 'X' } else { '-' });
+            }
+            internal.push('\n');
```
junniest, line 222: Printing newline characters manually is not a great idea because different OS can still use different newline chars. It's not critical anymore, but it's bad taste. I would recommend compiling the sequences into fasta records with the right gap patterns and then using `bio::io::fasta::Record` for output.

phylo/src/alignment/tests.rs
```diff
@@ -281,3 +335,156 @@ fn display_alignment() {
 
     assert_eq!(lines, true_lines);
 }
+
+#[test]
+fn build_ancestral_alignment_from_wrong_number_nodes() {
```
junniest, line 340: I found it difficult to understand what exactly are you testing here. But I think that you might need to refactor quite a bit of the actual code which will change the tests as well, so I will not comment on them for now.

junniest (commented):
Quite a lot of comments from me, sorry about that! But there's one point labelled with "important" that I would really like you to address before this can be merged. Some refactoring is also necessary to help make code more logical.

phylo/src/alignment/alignment_builder.rs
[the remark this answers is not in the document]
MattesMrzik: Thanks!

phylo/src/alignment/alignment_builder.rs
```diff
@@ -123,3 +126,134 @@ impl<'a> AlignmentBuilder<'a> {
...
+        let mut all_maps = leaf_maps;
+        // counter[node] keeps track of the sequence index during the procedurally generated mapping for the node
+        let mut counter = vec![0; n_nodes];
+        for site in 0..alignment_len {
+            // has_char[node] will be set to true if any leaf in the subtree rooted in node is not a gap
+            let mut has_char = vec![false; n_nodes];
+            // upward pass
+            for node in self.tree.postorder() {
+                match node {
+                    Leaf(id) => {
```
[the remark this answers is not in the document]
MattesMrzik: Okay, will change to idx

phylo/src/alignment/alignment_builder.rs
MattesMrzik: I will update the name when the new multiple ancestral seq alignment (masa) is implemented.

phylo/src/alignment/mod.rs
MattesMrzik: That makes much sense. thx

phylo/src/alignment/alignment_builder.rs
MattesMrzik: Makes sense. Thanks for pointing that out. Also, I will update this fn (split it into smaller pieces) and let it implement the ASR trait (that is not implemented yet).

MattesMrzik:
To summarize, what we have talked about: Implement a new trait msa/alignment, as well as a trait multiple ancestral seq alignment (masa) that implements everything that also msa does. Then the phylo_info has the member msa that must implement the msa trait. Since masa inherits msa the phylo_info also works with masa, which is necessary for tkf.

phylo/src/alignment/alignment_builder.rs
MattesMrzik: See #48

PR #50, opened 2025-04-16
fix: topo_optimiser starts from scratch for every prune because of in…

merlinio2000:
…correct tree update

As discussed with Julija over Teams:
Instead of applying multiple SPR moves per cycle (as is the idea of this project) the topo_optimiser ended up actually
only applying a single move each cycle.
This was because the local variable `tree` was only updated after each cycle, but used for every (prune,regraft) location to 
get the new SPR tree (which is supposed to be modified by the prune before but wasn't)


phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -52,7 +54,15 @@ impl<C: TreeSearchCost + Clone + Display> TopologyOptimiser<C> {
             tree = self.c.borrow().tree().clone();
             prev_cost = curr_cost;
 
-            for prune in &prune_locations {
+            let current_prunes = if cfg!(feature = "deterministic") {
+                possible_prunes.iter().collect()
```
junniest, line 58: @merlinio2000 Very much not sure this is a good way to do this, but my brain is not cooperating anymore.

phylo/src/optimisers/model_optimiser_tests.rs
```diff
@@ -481,7 +481,7 @@ fn pip_gtr_vs_gtr_params() {
 #[test]
 fn pip_protein_example() {
     let fldr = Path::new("./data/phyml_protein_example/");
-    let info = PIB::with_attrs(fldr.join("seqs.fasta"), fldr.join("true_tree.newick"))
+    let info = PIB::with_attrs(fldr.join("seqs.fasta"), fldr.join("example_tree.newick"))
```
junniest, line 484: This tree was not the true_tree it pretended to be, so it's now renamed.

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -91,6 +101,7 @@ impl<C: TreeSearchCost + Clone + Display> TopologyOptimiser<C> {
                     self.c
                         .borrow_mut()
                         .update_tree(best_tree, &[*prune, regraft]);
+                    tree = self.c.borrow().tree().clone();
```
junniest, line 104: Perfect fix and then I made it weird (:

junniest (commented):
@merlinio2000 please check that the feature I added is usable for benchmarks.

phylo/src/optimisers/topo_optimiser.rs
merlinio2000: Think this is the best way, actually forgot all about this macro

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -52,7 +60,13 @@ impl<C: TreeSearchCost + Clone + Display> TopologyOptimiser<C> {
             tree = self.c.borrow().tree().clone();
             prev_cost = curr_cost;
 
-            for prune in &prune_locations {
+            #[cfg(not(feature = "deterministic"))]
```
junniest, line 63: One question: why isn't this enclosed in a cfg_if macro? if the `rng` var is only defined inside the macro, then this could (potentially) use an undefined variable? It's not possible unless we e.g. end up with two different features instead of the one "deterministic", but my sense of symmetry and regularity isn't happy about this lol.

junniest (approved):
Looks good, just one comment on the `topo_optimiser.rs`, feel free to merge after!

phylo/src/optimisers/topo_optimiser.rs
merlinio2000: Ah I see the confusion. The difference between `#[cfg()]` and `cfg_if` is in scoping.
If I do (because #[cfg] only appies to the item below it, in this case the local block with two variables)
```
#[cfg()]
{
    let local = 0;
    let local2 = "a";
}
``` 

I cant access the variables further down because they were in a inner lexical scope.

But if I do
```
#[cfg()]
let local = 0;
#[cfg()]
let local2 = "a";
``` 

I can use these variables further down (given the feature is set) because they are not in a nested block/scope.

Because this gets tedious to write (especially with if/else type stuff) we have the cfg_if macro which essentially works by prepending #[cfg] to every line inside the ifs/elses. Functionally there is no difference between cfg_if and 'raw' #[cfg].



About the undefined variable: Im not sure I understand your worries, of course this is rust so it will not compile if there is any combination of features that is using rng and we forgot to define the variable for that feature.

junniest: Aaaah, ok, I see, sorry. Makes sense!

PR #52, opened 2025-04-24
Feature/use precomputed test results

junniest:
Added "use-precomputed" feature for local tests runs. Some long tests were relying on finding a file in a directory to decide whether the tree search needed to be rerun and there was no way reset that other than to delete the stored files. 

Now for local runs we can use `--features use-precomputed` to not rerun long tree searches if an existing tree is found. By default (without the feature flag) the tests will run the tree search as expected and will ignore any precomputed trees.

### DIFF
--- a/phylo/Cargo.toml
+++ b/phylo/Cargo.toml
@@ -8,7 +8,8 @@ edition = "2021"
 [features]
 ci_coverage = []
 deterministic = []
-use-precomputed = []
+# Feature that can speed up repeated test runs for when tree search is not affected and tests are ran locally.
+use-precomputed-test-results = []
 
 [profile.release]
 opt-level = 3
--- a/phylo/Cargo.toml
+++ b/phylo/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2021"
 [features]
 ci_coverage = []
 deterministic = []
-use_precomputed = []
+use-precomputed = []
 
 [profile.release]
 opt-level = 3
--- a/phylo/Cargo.toml
+++ b/phylo/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2021"
 [features]
 ci_coverage = []
 deterministic = []
+use_precomputed = []
 
 [profile.release]
 opt-level = 3
--- a/phylo/src/optimisers/topo_optimiser_tests.rs
+++ b/phylo/src/optimisers/topo_optimiser_tests.rs
@@ -4,6 +4,7 @@ use approx::assert_relative_eq;
 
 use crate::alignment::{Alignment, Sequences};
 use crate::evolutionary_models::FrequencyOptimisation::Empirical;
+#[cfg(feature = "use-precomputed")]
 use crate::io::write_newick_to_file;
 use crate::likelihood::TreeSearchCost;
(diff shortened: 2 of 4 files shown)


phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -267,264 +265,311 @@ fn pip_vs_subst_dna_tree() {
...
-            o.cost.info
+            let o = TopologyOptimiser::new(pip_cost.clone()).run().unwrap();
+            assert!(write_newick_to_file(&[o.cost.info.tree.clone()], pip_tree_file).is_ok());
+            (o.cost.info, o.final_cost)
         };
+    } else {
+        let o = TopologyOptimiser::new(pip_cost.clone()).run().unwrap();
+        let (pip_res, final_pip_logl) = (o.cost.info, o.final_cost);
+    }
+    }
```
merlinio2000, line 296: I think you could simplify all these cfg_if's into a single conditional compilation by creating a function like `cost_and_info....` which has two different definitions according to the feature.

```rust
#[cfg(feature="use-precomputed")] // or cfg_if probly also works
fn cost_and_info...(....) -> (PI, f64) {}
#[cfg(not(feature="use-precomputed"))]
fn cost_and_info...(....) -> (PI, f64) {}
```

and then simply call this function from your tests. I think you would need to repeat the function once for SCB instead of PIPCB

phylo/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2021"
 [features]
 ci_coverage = []
 deterministic = []
+use-precomputed = []
```
merlinio2000, line 11: maybe add 'test' somewhere to the name? If I were to use this library and discovered this feature flag in the docs page I would have completely wrong expectations as to what it does

junniest: Ah, good point! Shall I rename it to `test-use-precomputed`? I wonder if it's possible to feature gate the feature to only be available for tests? Although that might be too much inception.
I will also add a comment in the toml file.

phylo/src/optimisers/topo_optimiser_tests.rs
junniest: Good call. It was getting so repetitive that I phased out and didn't think to simplify.

phylo/Cargo.toml
merlinio2000: Sounds reasonable :+1: 
I **think** there even might be a way to do that but yeah...

phylo/src/optimisers/topo_optimiser_tests.rs
junniest: I tried to simplify now and just spent waaay too much time on it :D so I really should wrap it up now. But please let me know if it looks horrible.

merlinio2000:
Fascinating how you are unsure about some rust parts but keep producing macros just like that 😄

merlinio2000 (approved):

junniest:
> Fascinating how you are unsure about some rust parts but keep producing macros just like that 😄

But I can make chatGPT fix my macros (; And then it's all nice and weird :D And github Copilot is pissing me off, even though it's also chatGPT based...

PR #55, opened 2025-05-07
Feature/parsimony scores

junniest:


junniest:
The Dollo (PIP-like) parsimony score is ready, now I need to figure out how to include it in the tree search as an optional filter without overcomplicating everything.

phylo/src/parsimony/cost/dollo_parsimony_cost.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct DolloParsimonyInfo {
+    node_info: Vec<Vec<ParsimonySet>>,
+    node_info_valid: Vec<bool>,
+    node_insertion: Vec<FixedBitSet>,
+    leaf_sets: Vec<HashSet<usize>>,
+    node_leaf_sets: Vec<Vec<HashSet<usize>>>,
```
merlinio2000, line 180: These `Vec<Vec<...>>` seem like they could be more optimally (memory layout wise) managed through one continous, matrix style buffer. From the initialization code it looks like both dimensions are fixed

phylo/src/pip_model/mod.rs
```diff
@@ -486,12 +490,13 @@ impl<Q: QMatrix> PIPCost<Q> {
...
         let mut tmp = self.tmp.borrow_mut();
-        // TODO: unnecessary clone of whole matrix
-        let x_anc = tmp.anc[children[0]].clone();
-        let y_anc = tmp.anc[children[1]].clone();
-        tmp.anc[idx].set_column(1, &x_anc.column(0));
-        tmp.anc[idx].set_column(2, &y_anc.column(0));
-        tmp.anc[idx].set_column(0, &(x_anc.column(0) + y_anc.column(0)));
+
+        let x_anc = tmp.anc[children[0]].column(0).clone_owned();
+        let y_anc = tmp.anc[children[1]].column(0).clone_owned();
```
merlinio2000, line 495: I kinda wanted to measure eliminating clones in general in a follow up MR, so if you could make do with a possibly tiny bit slower binary for now would be very kind :)

phylo/src/parsimony/scoring/model_scoring.rs
```diff
@@ -129,11 +145,19 @@ impl ModelCosts {
...
 
-impl ParsimonyCosts for ModelCosts {
+impl<P: ParsimonyModel> ParsimonyScoring for ModelScoring<P> {
     fn r#match(&self, blen: f64, i: &u8, j: &u8) -> f64 {
         self.scoring(OrderedFloat(blen)).c[(self.alphabet.index(i), self.alphabet.index(j))]
     }
 
+    fn min_match(&self, blen: f64, i: &ParsimonySet, j: &ParsimonySet) -> f64 {
+        i.iter()
+            .zip(j.iter())
```
merlinio2000, line 155: not sure if this applies here but try to use zip_eq if the expectation is that both iterators are of equal length. Silent ignoring of elements can be a very hard to find error

phylo/src/parsimony/cost/dollo_parsimony_cost.rs
```diff
@@ -0,0 +1,257 @@
...
+    NodeIdx::{self, Internal, Leaf},
+    Tree,
+};
+use crate::Result;
+
+#[derive(Debug, Clone)]
+pub struct DolloParsimonyCost {
+    info: PhyloInfo,
+    tmp: RefCell<DolloParsimonyInfo>,
+    scoring: Box<dyn ParsimonyScoring>,
```
merlinio2000, line 24: is an enum/generic infeasible here? monomorphization can be a powerful optimization tool

phylo/src/parsimony/cost/basic_parsimony_cost.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    fn set_leaf(&self, node_idx: &NodeIdx) {
+        let node = self.info.tree.node(node_idx);
+        let idx = usize::from(node_idx);
+        if !self.tmp.borrow().node_info_valid[idx] {
+            if let Some(parent_idx) = node.parent {
+                self.tmp.borrow_mut().node_info_valid[usize::from(parent_idx)] = false;
+            }
+        }
+        self.tmp.borrow_mut().node_info_valid[idx] = true;
```
MattesMrzik, line 89: This could be moved into the `if !self.tmp.borrow().node_info_valid[idx] {`

phylo/src/parsimony/cost/dollo_parsimony_cost.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    fn set_internal(&self, node_idx: &NodeIdx) {
+        let node = self.info.tree.node(node_idx);
+        let idx = usize::from(node_idx);
+        let childx_idx = usize::from(&node.children[0]);
+        let childy_idx = usize::from(&node.children[1]);
+
+        let blen = node.blen;
+
+        if self.tmp.borrow().node_info_valid[idx] {
```
MattesMrzik, line 77: Maybe move this to after line 71, to avoid setting the unnecessary vars (childy_idx, ...) in the case where the node is valid.

phylo/src/parsimony/cost/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+    assert_eq!(cost.cost(), -4.0);
+
+    let tree_reroot = tree!("(((D:1,(B:1,A:0.5):0.5):1,C:2):0);");
+
+    let info_reroot = PhyloInfo {
+        msa: Alignment::from_aligned(seqs.clone(), &tree_reroot).unwrap(),
+        tree: tree_reroot,
+    };
+    let cost_reroot = DolloParsimonyCost::with_scoring(info_reroot, scoring).unwrap();
+    // This is a non-reversible cost function.
```
MattesMrzik, line 170: What do you mean by `non-reversible`?

MattesMrzik (approved):

phylo/src/pip_model/mod.rs
junniest: Haha, sure thing, will revert!

phylo/src/parsimony/scoring/model_scoring.rs
junniest: It doesn't apply but because it should have computed the scores for all to all pairs... This was the reason my scores kept making no sense.

phylo/src/parsimony/cost/dollo_parsimony_cost.rs
junniest: Ah, of course. Fixed.

phylo/src/parsimony/cost/dollo_parsimony_cost.rs
junniest: Fair!

phylo/src/parsimony/cost/tests.rs
junniest: If we reroot the tree, the cost may change.

phylo/src/parsimony/cost/dollo_parsimony_cost.rs
merlinio2000: you might not like to hear this but I usually hand roll stuff like this in like a 2DVec wrapper that simply stores Vec inside it and calculates the index by `i*(j+1) + j` (row-major storage). If you're more likely to access elements in column order you should flip that.

Judging by the lack of crates I could find through a quick search i guess most people are doing the same.
Guess storing non-primitives in a matrix is not a very non-math-coding thingy to do

junniest: Yeah, you're right, I don't like it :D I'll think about it!

phylo/src/optimisers/regraft_optimiser.rs
```diff
@@ -206,7 +199,7 @@ fn calc_spr_cost_with_blen_opt<C: TreeSearchCost + Clone + Display>(
     cost_func.update_tree(new_tree.clone(), &[prune_location, regraft]);
 
     let mut move_cost = cost_func.cost();
-    if move_cost <= base_cost {
+    if cost_func.blen_optimisation() && move_cost <= base_cost {
```
junniest, line 202: @maggimer I added this check to avoid trying to optimise branch lengths under a parsimony model. It can be easily ignored, the result of the tree search is the same, it just calls the score computation another bunch of times. I am not expecting anyone to run tree search under parsimony, but I am sure at some point someone inevitably will. But what I don't know is whether this one check that's being run every time is worth the time investment vs the time investment of calling the parsimony score too many times in an unlikely scenario...

PR #56, opened 2025-05-08
Feature/parallel topo variations

merlinio2000:
NOTE: DO-NOT-SQUASH
these 2 commits are intended to be measured independently

adds 2 iterations of different version of parallel regraft/topology optimisation

namely we search for the best scoring SPR move in parallel using rayon
- `par-regraft` simplest way possible using just par_iter and leaving everthing to rayon
- `par-regraft-chunk`  try to reduce thread overhead for smaller sets by running a fixed chunk per task
   this should reduce overhead if the unit of work is too small to warrant it
- `par-regraft-manual` a completely hand crafted forkjoin implementation using rayon that should in theory avoid heap allocations  (for tasks) almost entirely

### DIFF
--- a/phylo/Cargo.toml
+++ b/phylo/Cargo.toml
@@ -12,6 +12,7 @@ deterministic = []
 use-precomputed-test-results = []
 par-regraft = ["dep:rayon"]
 par-regraft-chunk = ["dep:rayon"]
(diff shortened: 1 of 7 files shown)


junniest (approved):
Looks good!

PR #60, opened 2025-05-26
Stochastic Tree Builder Class

### LINKED ISSUE #30: Move tree building with NJ in a builder class
Move the tree building functionality into a builder class out of the Tree class.

Switch the nj_matrices.rs:70 back to setting branch lengths to `0.0` instead of `f64::EPSILON` since this change was only because of the python calculation breaking.

kalxed:
Currently changes functionality so that you have to instantiate to use with methods. Not a consistent implementation with methods or associated functions yet, going to move to methods most likely. Distance function argument being developed right now.


phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,110 @@
...
+    NodeIdx, Tree,
+};
+
+use crate::alignment::Sequences;
+use crate::Result;
+use bio::alignment::distance::levenshtein;
+use nalgebra::{max, DMatrix};
+
+pub struct NJBuilder {
+    use_random: f64,
```
MattesMrzik, line 12: This var name sounds like a bool to me. Perhaps something like `temperature` is better. Then this also could be an `Option`, so if it is `None` we use the deterministic nj builder and if some(temperature) we use something like softmax with that temperature.

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -165,7 +165,10 @@ impl PhyloInfoBuilder {
             Some(tree_file) => self.read_tree(&sequences, tree_file)?,
             None => {
                 info!("Building NJ tree from sequences");
-                build_nj_tree(&sequences)?
+                let tree_builder = NJBuilder::new(0.00, None)?;
```
MattesMrzik: We could add a new field to `PhyloInfoBuilder` that is of generic type `TreeBuilder`, similar to how its done with the `aligner`.

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -165,7 +168,11 @@ impl PhyloInfoBuilder {
             Some(tree_file) => self.read_tree(&sequences, tree_file)?,
             None => {
                 info!("Building NJ tree from sequences");
-                build_nj_tree(&sequences)?
+                // Decide between default or new() usage, edit in future push
+                let _ = NJBuilder::new(None, None);
+                self.tree_builder
+                    .unwrap_or(Box::new(NJBuilder::default()))
```
kalxed: I'm unsure here whether this is the right way to do this...

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -6,16 +6,20 @@ use log::{info, warn};
...
+use crate::tree::nj_builder::{NJBuilder, Randomise};
+use crate::tree::tree_builder::TreeBuilder;
+use crate::tree::Tree;
 use crate::Result;
 
 pub struct PhyloInfoBuilder {
     sequence_file: PathBuf,
     tree_file: Option<PathBuf>,
     alignment_builder: Option<Box<dyn Aligner>>,
+    tree_builder: Option<Box<dyn TreeBuilder>>,
```
junniest: Ah, yes, I think that here having a Box is a good idea, generic would be too much with all the other stuff going on.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+use crate::tree::{NodeIdx, Tree};
+use crate::Result;
+use log::debug;
+use nalgebra::DMatrix;
+
+pub enum Randomise {
+    Deterministic,
+    Temperature(f64),
+}
+pub struct NJBuilder<D: EvolutionaryDistance> {
```
junniest: nitpick: newline before

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
+use crate::alignment::Sequences;
+use crate::evolutionary_distances::{EvolutionaryDistance, LevenshteinDNACorrected};
+use crate::tree::nj_matrices::{Mat, NJMat};
+use crate::tree::tree_builder::TreeBuilder;
+use crate::tree::{NodeIdx, Tree};
+use crate::Result;
+use log::debug;
+use nalgebra::DMatrix;
+
```
junniest, line 13: nitpick: sort use statements as discussed

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -144,7 +150,11 @@ impl PhyloInfoBuilder {
             Some(tree_file) => self.read_tree(&sequences, tree_file)?,
             None => {
                 info!("Building NJ tree from sequences");
-                build_nj_tree(&sequences)?
+                // Decide between default or new() usage, edit in future push
+                let _ = NJBuilder::new(Randomise::Deterministic, LevenshteinDNACorrected);
+                self.tree_builder
```
junniest: I think that default (second option) is the way to go here.

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -144,7 +150,11 @@ impl PhyloInfoBuilder {
...
+                    .unwrap_or(Box::new(NJBuilder::<LevenshteinDNACorrected>::default()))
+                    .build_tree(&sequences)?
```
junniest: Maybe a slightly better naming scheme would be `NJBuilder` -> `NJTreeBuilder` and `build_tree()` -> `build()`? Since otherwise it might seem like it can also build other things, potentially, but it's only ever capable of building trees...

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            );
+        };
+        Self {
+            temperature,
+            distance_function,
+        }
+    }
+
+    fn temperature_branch(&self, q: Mat) -> (usize, usize) {
+        debug_assert!(!q.is_empty(), "The input matrix must not be empty.");
```
junniest: I think that we don't need this check since the second check takes care of that.

phylo/src/evolutionary_distances/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+impl EvolutionaryDistance for LevenshteinDNACorrected {
+    fn dist(&self, a: &Record, b: &Record) -> f64 {
+        let seq_i = a.seq();
+        let seq_j = b.seq();
+        let dist = levenshtein(seq_i, seq_j) as f64;
+        let proportion_diff = f64::min(
+            dist / (max(seq_i.len(), seq_j.len()) as f64),
+            0.75 - f64::EPSILON,
+        );
+        // TODO: Doesn't work for proteins! @junniest
```
junniest: And now the name of the struct confirms that it, does not, in fact, make sense for proteins! :D

phylo/src/phylo_info/phyloinfo_builder.rs
junniest: and since it's being called essentially as `self.tree_builder.build_tree()` it feels like too much (:

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            temperature: Randomise::Deterministic,
+            distance_function: D::default(),
+        }
+    }
+}
+
+impl<D: EvolutionaryDistance> NJBuilder<D> {
+    // With the way we are using Option makes sense to use Into to allow for passing
+    pub(crate) fn new(temperature: Randomise, distance_function: D) -> Self {
+        // Add checking of unwrap for default when not None is passed in? Warning to user
```
junniest: I don't think this comment is relevant anymore? If Randomise::Temperature is selected, the actual value will always be an f64, so no need to worry about that.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+        NJBuilder {
+            temperature: Randomise::Deterministic,
+            distance_function: D::default(),
+        }
+    }
+}
+
+impl<D: EvolutionaryDistance> NJBuilder<D> {
+    // With the way we are using Option makes sense to use Into to allow for passing
+    pub(crate) fn new(temperature: Randomise, distance_function: D) -> Self {
```
junniest: I think this should be `pub`.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+        // Add checking of unwrap for default when not None is passed in? Warning to user
+        // None temperature should be allowed, or if float, add stochastic
+        if let Randomise::Temperature(t) = temperature {
+            assert!(
```
junniest: A regular `assert!` here will cause an unexpected panic, although e.g. 1.5 would be a valid input based on the input type. I don't think we should use `assert!` in production code, since it is like an uncatchable exception.

Do you think that an incorrect temperature value (t < 0 or t > 1) is a failure scenario? I.e. if the value is not what you expect, should the creation fail? If so, I think the signature should be `new(...) -> Result<Self>`. But it seems excessive  to fail when we get a technically allowed value, and then we'd have to handle that failure in multiple places. I'd say that the easiest way to handle this is how it's done, for example, in `substitution_models/dna_models/mod.rs:15`, use `warn!` to print the warning message and then set the temperature to an allowed value, e.g. to 0.0 is t < 0.0 and to 1.0 if it's > 1.0.

What do you think?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    }
+
+    fn temperature_branch(&self, q: Mat) -> (usize, usize) {
+        debug_assert!(!q.is_empty(), "The input matrix must not be empty.");
+        debug_assert!(
+            q.ncols() > 1 && q.nrows() > 1,
+            "The input matrix should have more than 1 element."
+        );
+        debug_assert!(q.ncols() == q.nrows(), "The input matrix should be square.");
+        println!("{q}");
```
junniest: nitpick: Rogue print statement (:

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+        // Test better outputs of computer_nj_q() as this affects this...
+        // Based on nj_correct test case, negative distances are here. Otherwise, multiplying by negative 1 should correctly make large distances very small, and small distances, still small, or large if negative
```
junniest: As we discussed yesterday, we now know that both positive and negative values are possible, so this comment is obsolete.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+                "Temperature must be between 0.0 and 1.0 inclusive"
+            );
+        };
+        Self {
+            temperature,
+            distance_function,
+        }
+    }
+
+    fn temperature_branch(&self, q: Mat) -> (usize, usize) {
```
junniest: Let's rename the stupid q matrix to something more meaningful now (: Maybe `expected_length_diff`?.. That's too long, but the matrix represents expected differences in total tree length if we merge a given pair, so something along those lines?..

phylo/src/tree/nj_builder.rs
junniest: Although if we go through with the idea of using a vector instead of an unnecessary matrix, these checks will become unnecessary.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+        );
+        debug_assert!(q.ncols() == q.nrows(), "The input matrix should be square.");
+        println!("{q}");
+        // Test better outputs of computer_nj_q() as this affects this...
+        // Based on nj_correct test case, negative distances are here. Otherwise, multiplying by negative 1 should correctly make large distances very small, and small distances, still small, or large if negative
+        let mut exp_mat = q.scale(-1.0);
+        // e^i on all elements of matrix
+        exp_mat = exp_mat.lower_triangle().map(|i| i.exp());
+        // With large matrices/numbers there is a slight discrepancy due to floating point numbers, greater than 1
+        let exp_sum = exp_mat.lower_triangle().sum() - exp_mat.diagonal().sum();
```
junniest: I think that this will be simplified away since we agreed to not use a matrix for this but rather a vector?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            exp_mat.lower_triangle().sum() - exp_mat.diagonal().sum()
+        );
+
+        // Use temperature with uniform value to calculate probability
+        let uniform: f64 = 1.0 / (((q.nrows().pow(2) - q.nrows()) / 2) as f64);
+        // Temperature probabilities, temp=0.0 means uniform, temp=1.0 means softmax of distances
+        let temperature = match self.temperature {
+            Randomise::Temperature(f64) => f64,
+            Randomise::Deterministic => {
+                panic!("Called temperature_branch while in Deterministic mode")
```
junniest: There should be a better way of handling this. We should not resort to panicking, especially since this is a private method and we are the only ones who can call it (; Potentially, this can be called in `build_nj_tree_from_matrix` (also a horrible name inherited from my code) with the temperature passed as a parameter, so there is absolutely no chance that it will be called without a valid value and this match will go away. I think this would be fine since the match happens in the caller method anyway.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            }
+        };
+        exp_mat = exp_mat
+            .lower_triangle()
+            .map(|i| (temperature * uniform) + (1.0 - temperature * i));
+        // This is our criterion for choosing, can use random seed because iteration is deterministic
+        let branch_choice: f64 = rand::random();
+        let mut prob_sum = 0.0;
+        for i in 1..q.nrows() {
```
junniest: This will be a lot simpler (and more generic) once we switch to a vector instead of the stupid matrix.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            for j in 0..i {
+                let val = exp_mat[(i, j)];
+                prob_sum += val;
+                if prob_sum > branch_choice {
+                    return (i, j);
+                }
+            }
+        }
+        // Change? default, warning, fallback to argmin, this should never happen based on probability contraints.
+        panic!("No return from softmax branch");
```
junniest: Hm, could this happen due to numerical error? In that case the answer is simply the last entry, isn't it? So it probably should just have `return(q.nrows()-1, q.nrows()-1)` at the very end, or, if you think that's not a valid scenario, `unreachable!()`.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+        let n = nj_data.distances.ncols();
+        let mut tree = Tree::new(sequences)?;
+        let root_idx = usize::from(&tree.root);
+        //Wrapped in temperature check for now, can implement stochastic better in future
+        for cur_idx in n..=root_idx {
+            let q = nj_data.compute_nj_q();
+            // This is where we choose our next branch, if temperature exists softmax, if not argmin, default behavior
+            let (i, j) = match self.temperature {
+                Randomise::Temperature(_t) => self.temperature_branch(q),
+                Randomise::Deterministic => NJBuilder::<D>::argmin_wo_diagonal(q),
```
junniest: This should probably also become a `self.argmin_wo_diagonal` method. But I hope that you can do away with the stupid implementation altogether (:

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+                .remove_merged_nodes(i, j);
+        }
+        tree.n = n;
+        tree.complete = true;
+        tree.compute_postorder();
+        tree.compute_preorder();
+        tree.length = tree.nodes.iter().map(|node| node.blen).sum();
+        Ok(tree)
+    }
+    //Converted this to method instead of associated function, we can decide which to use
```
junniest: There's some guidelines on this, but I am not sure what I think about methods vs associated functions when the body of the code doesn't really need self...

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+        for i in 0..nseqs {
+            for j in (i + 1)..nseqs {
+                let corrected_dist = self
+                    .distance_function
+                    .dist(sequences.record(i), sequences.record(j));
+                distances[(i, j)] = corrected_dist;
+                distances[(j, i)] = corrected_dist;
+            }
+        }
+        NJMat {
```
junniest: Also what an atrocious name for a struct, `NJMat`. Renaming ideas are welcome.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            idx: (0..nseqs).map(NodeIdx::Leaf).collect(),
+            distances,
+        }
+    }
+}
+// Implement the tests at the bottom of this module for ability to use private functions
+#[cfg(test)]
+mod private_tests {
+    //From test.rs in tree, so we can use macros
```
junniest: nitpick: space after `//`

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    use crate::{record_wo_desc as record, tree};
+    use nalgebra::{dmatrix, DMatrix};
```
junniest: nitpick: sort use statements

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+        // Have to wrap distance function in Some, because otherwise Into has to be implemented for DistanceFunction type
+        let _nj_builder = NJBuilder::new(Randomise::Temperature(0.0), LevenshteinDNACorrected {});
+        let _nj_builder = NJBuilder::new(Randomise::Temperature(1.0), LevenshteinDNACorrected {});
+    }
+
+    #[test]
+    #[should_panic]
+    fn nj_builder_panic_creation() {
+        //temperature over 1.0
+        let _nj_builder = NJBuilder::new(Randomise::Temperature(2.0), LevenshteinDNACorrected {});
```
junniest: I still think that this creation should not panic, but rather set temperature to 1.0 and move on (:

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+
+        assert_eq!(nj_tree.root, I(6));
+        assert_eq!(nj_tree.nodes, nodes);
+    }
+
+    #[test]
+    fn nj_builder_correct_creation() {
+        let _nj_builder: NJBuilder<LevenshteinDNACorrected> = NJBuilder::default();
+        let _nj_builder = NJBuilder::new(Randomise::Deterministic, LevenshteinDNACorrected {});
+        // Have to wrap distance function in Some, because otherwise Into has to be implemented for DistanceFunction type
```
junniest: Hm, is this an older comment? I don't seem to understand what you mean here.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    //         Node::new_internal(5, Some(I(6)), vec![L(3), L(2)], 1.0, "".to_string()),
+    //         Node::new_internal(6, None, vec![I(4), I(5)], 0.0, "".to_string()),
+    //     ];
+    //     assert_eq!(nj_tree.root, I(6));
+    //     assert_eq!(nj_tree.nodes, nodes);
+    // }
+
+    // #[test]
+    // fn nj_builder_softmax() {
+    //     // Stolen from nj_correct test
```
junniest: Definitely not stolen, just used (;

phylo/src/tree/tree_builder.rs
```diff
@@ -0,0 +1,7 @@
+use crate::alignment::Sequences;
+use crate::tree::Tree;
+use crate::Result;
+
+pub trait TreeBuilder {
+    fn build_tree(&self, seqs: &Sequences) -> Result<Tree>;
```
junniest: Yep, I think this should be simply `build()`.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    }
+
+    #[test]
+    #[should_panic]
+    fn nj_builder_panic_creation() {
+        //temperature over 1.0
+        let _nj_builder = NJBuilder::new(Randomise::Temperature(2.0), LevenshteinDNACorrected {});
+    }
+
+    // // Rethink adding these tests, would need random seed to work
```
junniest: Hopefully I'll implement the global random provider within the next few days, stay tuned!

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    //         Node::new_leaf(2, Some(I(5)), 2.0, "C2".to_string()),
+    //         Node::new_leaf(3, Some(I(5)), 7.0, "D3".to_string()),
+    //         Node::new_internal(4, Some(I(6)), vec![L(0), L(1)], 1.0, "".to_string()),
+    //         Node::new_internal(5, Some(I(6)), vec![L(3), L(2)], 1.0, "".to_string()),
+    //         Node::new_internal(6, None, vec![I(4), I(5)], 0.0, "".to_string()),
+    //     ];
+    //     assert_eq!(nj_tree.root, I(6));
+    //     assert_eq!(nj_tree.nodes, nodes);
+    // }
+}
```
junniest, line 767: Sorry, I said that we shouldn't add more distance measures, but we should probably add wrap basic Levenstein edit distance in a DistanceMeasure and add a test that goes through the whole pipeline, as it would be used in code, so that `nj_builder.build_tree()` is used directly with sequences.

junniest (commented):
Good stuff (; It's a lot of comments, sorry about that, but it's mostly for discussion, not because something is wrong.

phylo/src/phylo_info/phyloinfo_builder.rs
kalxed: Yes I think this was practically mirrored from the alignment_builder as Mattes pointed out

phylo/src/phylo_info/phyloinfo_builder.rs
kalxed: Haha good point will change

kalxed: I think the only point actually that comes up is if a TreeBuilder should ever be able to build something other than a tree, you could implement multiple builder traits to one type. Then in this case you might want a difference between `build_tree()` and `build_alignment` or something? In the end I think it would work out to just `self.build()` so if there are other builder traits it could just collide in that sense. Let me know what you think of this, but I'm making the change because semantically it makes sense to me.

phylo/src/tree/nj_builder.rs
kalxed: Yes... I will remove my old comments now

phylo/src/tree/nj_builder.rs
kalxed: I think PhyloInfo might use the build() direcdtly but it makes sense to add a test within the private tests to do that too.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            distance_function: D::default(),
+        }
+    }
+}
+
+impl<D: EvolutionaryDistance> NJBuilder<D> {
+    // With the way we are using Option makes sense to use Into to allow for passing
+    pub(crate) fn new(temperature: Randomise, distance_function: D) -> Self {
+        // Add checking of unwrap for default when not None is passed in? Warning to user
+        // None temperature should be allowed, or if float, add stochastic
```
kalxed: I believe this should be pub not pub(crate) as it can be called outside of the crate?

phylo/src/tree/nj_builder.rs
kalxed: Depends if we want to be able to pass the q matrix directly in, unless your idea is to rework the "Q" matrix method as well which looking at it would make sense. Regardless we have to convert between a distance matrix and vector at some point, but we can technically save on computations just by converting to a vector (or at least not do things twice)

phylo/src/tree/nj_builder.rs
kalxed: Didn't mean to be there :O

phylo/src/tree/nj_builder.rs
kalxed: Yes, depends where the conversion to vector happens but will sort out the functionality anyways

phylo/src/tree/nj_builder.rs
kalxed: Will do this solution even though it's sad to see `self` go away, makes more sense for testing anyways

phylo/src/tree/nj_builder.rs
kalxed: Ah yes that makes sense to me, good catch

phylo/src/phylo_info/phyloinfo_builder.rs
junniest: I don't think that something like `NJTreeBuilder` should ever implement functionality for building an entirely different object, but even when two traits are in scope and have the same method name it's easily resolved. If only one is implemented for the struct it's not an issue at all, but otherwise there's a way to specify which trait you want for each call. For example, I was testing to see that the outputs of the models when used as `ModelSearchCost` and as `TreeSearchCost` are the same (`likelihood/tests.rs:16`):
```
#[cfg(test)]
fn search_costs_equal_template<C: ModelSearchCost + TreeSearchCost>(cost: C) {
    assert_eq!(ModelSearchCost::cost(&cost), TreeSearchCost::cost(&cost));
}
```
It's a bit bulky, but that's one way to work around same method names in traits.

phylo/src/tree/nj_builder.rs
junniest: The distance matrix can remain a matrix, or also be converted into a vector, I think this does not matter too much. Keeping it as a matrix might make it easier to access the right elements and we don't need to do any argmin/max over it.

But the Q matrix, so the expected tree length change matrix, can just turn into a vector everywhere. The only use for it is to find the next pair of nodes to join, and having it in a vector I think will make all the softmax operations on it a lot easier.

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -144,7 +150,10 @@ impl PhyloInfoBuilder {
             Some(tree_file) => self.read_tree(&sequences, tree_file)?,
             None => {
                 info!("Building NJ tree from sequences");
-                build_nj_tree(&sequences)?
+                // Decide between default or new() usage, edit in future push
```
junniest: nitpick: obsolete comment

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+use crate::tree::{NodeIdx, Tree};
+use crate::Result;
+
+pub enum Randomise {
+    Deterministic,
+    Temperature(f64),
+}
+
+pub struct NJTreeBuilder<D: EvolutionaryDistance> {
+    temperature: Randomise,
```
junniest: I don't really like this name for the field, it's very difficult to read like this. What does "temperature: randomise" mean? (: I also got confused somewhere later about why the temperature is being used in a match statement in a strange way. I would suggest renaming this to even just `randomise` to reduce confusion, although I'm sure there's a better name for it.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+    distance_function: D,
+}
+
+impl<D: EvolutionaryDistance> TreeBuilder for NJTreeBuilder<D> {
+    fn build(&self, sequences: &Sequences) -> Result<Tree> {
+        let nj_data = self.compute_distance_matrix(sequences);
```
junniest: I remember that this is a moved piece of code, but the naming is atrocious again. What the hell is `nj_data`, it's a damn distance matrix, could we rename this to `distances`?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+            Randomise::Temperature(t) if t < 0.0 => {
+                debug!("Temperature shouldn't be less than 0.00, setting to 0.00");
+                Randomise::Temperature(0.0)
+            }
+            Randomise::Temperature(t) if t > 1.0 => {
+                debug!("Temperature shouldn't be greater than 1.00, setting to 1.00");
+                Randomise::Temperature(1.0)
+            }
+            _ => temperature,
+        };
```
junniest: I really find this piece difficult to read and understand. I think that `temperature` should not be called that because it has a double meaning which is confusing. I'd also suggest splitting the `new` method into two, basic `new` without a temperature which defaults to Strategy::Deterministic, and `with_temperature` or `with_softmax` which accepts an f64 (or maybe a newtype called `Temperature`).

Then you can use `temperature.clamp(0.0, 1.0)` to restrict the values in `with_temperature`.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+use rand::SeedableRng;
+
+use crate::alignment::Sequences;
+use crate::evolutionary_distances::{EvolutionaryDistance, LevenshteinDNACorrected};
+use crate::tree::nj_matrices::NJMat;
+use crate::tree::tree_builder::TreeBuilder;
+use crate::tree::{NodeIdx, Tree};
+use crate::Result;
+
+pub enum Randomise {
```
junniest: I think a better naming scheme would be to call this `Strategy` and have the two variants `Deterministic` and `SoftMax(f64)`. Better yet, let's declare a new type for Temperature. So it could look something like this:
```
struct Temperature(f64);

pub enum Strategy {
    Deterministic,
    SoftMax(Temperature),
}
```

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+            }
+            _ => temperature,
+        };
+        Self {
+            temperature,
+            distance_function,
+        }
+    }
+
+    fn index_to_ij_rowwise_nodiag(k: usize) -> (usize, usize) {
```
junniest: maybe call this `lower_triag_index`? I think that in general we do row-wise operations so this part of the name is somewhat redundant, and the nodiag would be covered by indicating that it's the lower triangle.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        let p = ((1 + 8 * k).isqrt() - 1) / 2;
+        let i = p + 1;
+        let j = k - p * (p + 1) / 2;
+        (i, j)
+    }
+
+    fn softmax_vector_mut(mut v: DVector<f64>) -> DVector<f64> {
+        v = v.map(|i| i.exp());
+        let sum_j = v.sum();
+        v.unscale(sum_j)
```
junniest: Maybe it's worth adding a `debug_assert!` to check that the vector (roughly) sums up to 1 now?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+
+    fn temperature_branch(delta_tree_len: DVector<f64>, temperature: f64) -> (usize, usize) {
+        debug_assert!(
+            !delta_tree_len.is_empty(),
+            "The input vector must not be empty."
+        );
+        if delta_tree_len.len() == 1 {
+            return NJTreeBuilder::<LevenshteinDNACorrected>::index_to_ij_rowwise_nodiag(0);
+        }
+        // Based on nj_correct test case, negative distances are common in compute_nj_q, which should be multiplied by -1 for small (desireable) distances to be larger in comparison to large distances. Allows for use of softmax
```
junniest: This comment is outdated (because thankfully `compute_nj_q`) doesn't exist anymore, but I am not sure I am reading this correctly anyway. It sounds as if you are suggesting to multiply parts of the matrix (small entries), which I can see is not what is happening anyway, but it confused me (:
Also, softmax can handle both positive and negative numbers meaningfully, right? Because we just want to prefer the smallest negative number as the join that will give us the highest improvement in tree length.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        let scaled_mat = delta_tree_len.scale(-1.0);
```
junniest: Maybe a better name could be `inverted_delta`? Since the scaling is just by -1.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        // I think this messes up scaled_mat because of function implementation, borrows and never gives it back, could be reworked I think?
+        let mut exp_mat = NJTreeBuilder::<LevenshteinDNACorrected>::softmax_vector_mut(scaled_mat);
```
junniest: This should definitely be
```
let mut exp_mat = NJTreeBuilder::<D>::softmax_vector_mut(scaled_mat);
```

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        debug_assert!(
+            !delta_tree_len.is_empty(),
+            "The input vector must not be empty."
+        );
+        if delta_tree_len.len() == 1 {
+            return NJTreeBuilder::<LevenshteinDNACorrected>::index_to_ij_rowwise_nodiag(0);
+        }
+        // Based on nj_correct test case, negative distances are common in compute_nj_q, which should be multiplied by -1 for small (desireable) distances to be larger in comparison to large distances. Allows for use of softmax
+        let scaled_mat = delta_tree_len.scale(-1.0);
+        // I think this messes up scaled_mat because of function implementation, borrows and never gives it back, could be reworked I think?
```
junniest: It doesn't borrow, it takes a `mut DVector`, but then on the last step it copies the vector instead of modifying in place.

phylo/src/tree/nj_builder.rs
junniest: `v.unscale(sum_j)` should be `v.unscale_mut(sum_j)`

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        let mut exp_mat = NJTreeBuilder::<LevenshteinDNACorrected>::softmax_vector_mut(scaled_mat);
+        debug!("Probability Sum: {}", exp_mat.sum());
```
junniest: I think it's best to use `debug!` and other output for meaningful messages to the user, e.g. for warnings when temperature is a weird value. Someone seeing this in a debug log will not be able to interpret what this means, so I think it's best to replace it with a `debug_assert!()` in `softmax_vector_mut()` instead.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        // Use temperature with uniform value to calculate probability
+        let uniform: f64 =
```
junniest: I think that type annotation here isn't necessary.
Also I don't think I understand what you meant in the comment :D

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+            1.0 / (((delta_tree_len.nrows().pow(2) - delta_tree_len.nrows()) / 2) as f64);
+        // Temperature probabilities, temp=0.0 means uniform, temp=1.0 means softmax of distances
+        exp_mat = exp_mat.map(|i| (temperature * uniform) + ((1.0 - temperature) * i));
+        // This is our criterion for choosing, can use random seed because iteration is deterministic
+        let dist = WeightedIndex::new(exp_mat.data.as_vec().iter()).unwrap();
+        let mut rng = StdRng::from_seed([0; 32]);
+        NJTreeBuilder::<LevenshteinDNACorrected>::index_to_ij_rowwise_nodiag(dist.sample(&mut rng))
```
junniest: Again, should be
```
NJTreeBuilder::<D>::index_to_ij_rowwise_nodiag(dist.sample(&mut rng))
```
or `Self::index_to_ij...`.

Hm, if it is deterministic, why does WeightedIndex need rng in the first place? This is very confusing.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        for i in 0..q.nrows() {
+            let val = &q[i];
+            if val < val_min {
+                val_min = val;
+                // Convert usize to u32?
+                arg_min =
+                    vec![NJTreeBuilder::<LevenshteinDNACorrected>::index_to_ij_rowwise_nodiag(i)];
+            } else if val == val_min {
+                arg_min
+                    .push(NJTreeBuilder::<LevenshteinDNACorrected>::index_to_ij_rowwise_nodiag(i));
```
junniest: Here too, `NJTreeBuilder<LevenshteinDNACorrected>` -> `NJTreeBuilder<D>`.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+    fn build_nj_tree_from_matrix(&self, mut nj_data: NJMat, sequences: &Sequences) -> Result<Tree> {
+        let n = nj_data.distances.ncols();
+        let mut tree = Tree::new(sequences)?;
+        let root_idx = usize::from(&tree.root);
+        for cur_idx in n..=root_idx {
+            let q = nj_data.compute_nj_delta_tree_length();
+            let (i, j) = match self.temperature {
+                Randomise::Temperature(t) => Self::temperature_branch(q, t),
+                Randomise::Deterministic => {
+                    NJTreeBuilder::<LevenshteinDNACorrected>::argmin_wo_diagonal(q)
```
junniest: -> `Self::argmin_wo_diagonal(q)`

phylo/src/tree/nj_builder.rs
junniest: or maybe even
```
let mut exp_mat = Self::softmax_vector_mut(scaled_mat);
```

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        if q.nrows() == 1 {
+            return (1, 0);
+        }
+        let mut arg_min = vec![];
+        let mut val_min = &f64::MAX;
+        for i in 0..q.nrows() {
+            let val = &q[i];
+            if val < val_min {
+                val_min = val;
+                // Convert usize to u32?
```
junniest: why?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        // Temperature probabilities, temp=0.0 means uniform, temp=1.0 means softmax of distances
+        exp_mat = exp_mat.map(|i| (temperature * uniform) + ((1.0 - temperature) * i));
+        // This is our criterion for choosing, can use random seed because iteration is deterministic
+        let dist = WeightedIndex::new(exp_mat.data.as_vec().iter()).unwrap();
+        let mut rng = StdRng::from_seed([0; 32]);
+        NJTreeBuilder::<LevenshteinDNACorrected>::index_to_ij_rowwise_nodiag(dist.sample(&mut rng))
+    }
+
+    // Used with Randomise::Deterministic
+    fn argmin_wo_diagonal(q: DVector<f64>) -> (usize, usize) {
```
junniest: Also now the name makes no sense (: and input-output are confusing, since it's taking a vector and returning a matrix index tuple for some reason. I'd say that both `argmin` (we can ditch the `wo_dioagonal`) and `temperature_branch` should return an index within the vector andthen be transformed into the matrix index in the caller. This also would reduce the call stack.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        let root_idx = usize::from(&tree.root);
+        for cur_idx in n..=root_idx {
+            let q = nj_data.compute_nj_delta_tree_length();
+            let (i, j) = match self.temperature {
+                Randomise::Temperature(t) => Self::temperature_branch(q, t),
+                Randomise::Deterministic => {
+                    NJTreeBuilder::<LevenshteinDNACorrected>::argmin_wo_diagonal(q)
+                }
+            };
+            println!("{i}, {j}");
```
junniest: Rogue print statement

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+            let idx_new = cur_idx;
+            let (blen_i, blen_j) = nj_data.branch_lengths(i, j, cur_idx == root_idx);
+            tree.add_parent(idx_new, &nj_data.idx[i], &nj_data.idx[j], blen_i, blen_j);
+            nj_data = nj_data
```
junniest: please, let's rename this to something more meaningful, `node_distances` maybe?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        tree.length = tree.nodes.iter().map(|node| node.blen).sum();
+        Ok(tree)
+    }
+
+    fn compute_distance_matrix(&self, sequences: &Sequences) -> NJMat {
+        let nseqs = sequences.len();
+        let mut distances = DMatrix::zeros(nseqs, nseqs);
+        for i in 0..nseqs {
+            for j in (i + 1)..nseqs {
+                let corrected_dist = self
```
junniest: This can now just be `dist` because we don't know if the distance function does correction.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+    #[cfg(test)]
+    fn is_unique<T: std::cmp::Eq + std::hash::Hash>(vec: &[T]) -> bool {
+        let set: std::collections::HashSet<_> = vec.iter().collect();
+        set.len() == vec.len()
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_argmin_fail() {
+        //Changed test for empty vector, since argmin with 1 vector length will return first branch as delta_tree_length vector is different
```
junniest: Hm, but this will only panic in debug because of the assertion. Do we really want it to panic or maybe just return a 0?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+
+    #[test]
+    fn compute_distance_matrix_far() {
+        let sequences = Sequences::new(vec![
+            record!("A0", b"AAAAAAAAAAAAAAAAAAAA"),
+            record!("B1", b"AAAAAAAAAAAAAAAAAAAA"),
+            record!("C2", b"AAAAAAAAAAAAAAAAAAAAAAAAA"),
+            record!("D3", b"CAAAAAAAAAAAAAAAAAAA"),
+        ]);
+        //For now, may instantiate NJBuilder instance every time
```
junniest: Why for now?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        }
+        }
+    }
+
+    #[cfg(not(feature = "deterministic"))]
+    fn rng_len(l: usize) -> usize {
+        rand::random::<usize>() % l
+    }
+
+    fn build_nj_tree_from_matrix(&self, mut nj_data: NJMat, sequences: &Sequences) -> Result<Tree> {
```
junniest: I think this name is now obsolete too, since there's no other thing that `NJTreeBuilder` can build, so maybe rename to `build_from_distances()`?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+            Node::new_internal(5, Some(I(6)), vec![L(3), L(2)], 1.0, "".to_string()),
+            Node::new_internal(6, None, vec![I(4), I(5)], 0.0, "".to_string()),
+        ];
+
+        assert_eq!(nj_tree.root, I(6));
+        assert_eq!(nj_tree.nodes, nodes);
+    }
+
+    #[test]
+    fn nj_builder_correct_creation() {
```
junniest, line 522: I'd split these up into subcategories, like `correct_default`, `correct_with_temp`, etc, and then once it is created check that the internal fields are actually set to what you expect. So, for example, if you try to create a builder with a temperature of `-5.0` you should end up with `t = 0.0`, but if you create it with `0.75` that should be set correctly.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        let nj_distances = NJMat {
+            idx: (0..5).map(NodeIdx::Leaf).collect(),
+            distances: dmatrix![
+                0.0, 5.0, 9.0, 9.0, 8.0;
+                5.0, 0.0, 10.0, 10.0, 9.0;
+                9.0, 10.0, 0.0, 8.0, 7.0;
+                9.0, 10.0, 8.0, 0.0, 3.0;
+                8.0, 9.0, 7.0, 3.0, 0.0],
+        };
+        let q = nj_distances.compute_nj_delta_tree_length();
```
junniest: I think this method should also get renamed, why do we keep repeating that it's `nj`, it's in the `NJBuilder` module lol. Can just be `delta_tree_length` or something.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+            (1, 0)
+        );
+    }
+
+    #[test]
+    fn softmax_vector_mut_test() {
+        let delta_tree_length = dvector![1.3, 5.1, 2.2, 0.7, 1.1];
+        let softmax_vector =
+            NJTreeBuilder::<LevenshteinDNACorrected>::softmax_vector_mut(delta_tree_length);
+        println!("{softmax_vector:?}");
```
junniest: rogue println

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+
+    fn index_to_ij_rowwise_nodiag(k: usize) -> (usize, usize) {
+        // 0 indexed
+        let p = ((1 + 8 * k).isqrt() - 1) / 2;
+        let i = p + 1;
+        let j = k - p * (p + 1) / 2;
+        (i, j)
+    }
+
+    fn softmax_vector_mut(mut v: DVector<f64>) -> DVector<f64> {
```
junniest: I think that this doesn't need `vector` in the name, since this is the only type the func will accept.

phylo/src/tree/nj_matrices.rs
```diff
@@ -72,15 +72,15 @@ impl NJMat {
...
-            } else {
-                (n - 2) as f64 * self.distances[(r, c)] - s[r] - s[c]
+        let mut delta_tree_len = Vec::with_capacity(n * (n + 1) / 2 - n);
+        for r in 1..n {
+            for c in 0..r {
+                delta_tree_len.push((n - 2) as f64 * self.distances[(r, c)] - s[r] - s[c])
             }
-        })
+        }
+        DVector::from_vec(delta_tree_len)
```
junniest: You could probably make a DVector of a fixed size from the start and populate that, since we know exactly how many entries there will be. Then there will be no need for a Vec allocation and then conversion.

phylo/src/tree/tests.rs
```diff
@@ -897,3 +690,4 @@ fn rf_distance_against_raxml() {
     assert_eq!(tree_phyml.robinson_foulds(tree_from_nj), 0);
     assert_eq!(tree.robinson_foulds(tree_from_nj), 0);
 }
+//Should probably add a test for build_nj_tree(), though as long as composite functions work, I doubt there is an issue...?
```
junniest: Obsolete comment

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        let delta_tree_length =
+            dvector![-50.0, -38.0, -38.0, -34.0, -34.0, -40.0, -34.0, -34.0, -40.0, -48.0];
+        assert_eq!(
+            NJTreeBuilder::<LevenshteinDNACorrected>::argmin_wo_diagonal(delta_tree_length),
+            (1, 0)
+        );
+    }
+
+    #[test]
+    fn softmax_vector_mut_test() {
```
junniest: We don't need to add `test_` or `_test` to test funcs (: I also keep doing that, but if you look at the test names, it's often `module::test::test_something` and the second `test` is really unnecessary.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,602 @@
...
+        };
+        let q = nj_distances.compute_nj_delta_tree_length();
+        assert_eq!(
+            q,
+            dvector![-50.0, -38.0, -38.0, -34.0, -34.0, -40.0, -34.0, -34.0, -40.0, -48.0]
+        )
+    }
+
+    #[test]
+    fn argmin_wo_diagonal_vector() {
```
junniest: For this we could probably add more edge cases, e.g. when all the entries in the vector are the same, we should get the value from the fake random generator, since for multiple entries we pick uniformly at random.

phylo/src/tree/nj_builder.rs
kalxed: Yes it can handle both meaningfully, we flip the sign off all numbers to get minimum instead in the softmax_uniform_to_lower_triangle

phylo/src/tree/nj_builder.rs
kalxed: That comment looks like nonsense

phylo/src/tree/nj_builder.rs
kalxed: Thank you!

phylo/src/tree/nj_builder.rs
kalxed: I think it makes sense to panic because if you return 0 that would seem like you can access that index of the vector

phylo/src/tree/nj_builder.rs
kalxed: I like that, I think the return type is only possibly confusing but since it's internal should be good

phylo/src/tree/nj_builder.rs
kalxed: Fixed, is it possible to be less verbose in the tests?

phylo/src/tree/nj_builder.rs
kalxed: Comment was old and updated both to match.

phylo/src/tree/nj_builder.rs
kalxed: I think this is fixed, `argmin` and `softmax_uniform` for index, `softmax` for softmax vector returned

phylo/src/tree/nj_builder.rs
kalxed: I called it `distances` but `node_distances` may be better...

phylo/src/tree/tree_builder.rs
```diff
@@ -0,0 +1,7 @@
+use crate::alignment::Sequences;
+use crate::tree::Tree;
+use crate::Result;
+
+pub trait TreeBuilder {
```
MattesMrzik, line 9: Even though this is self-explanatory, could you add a doc here?

phylo/src/tree/nj_matrices.rs
```diff
@@ -72,15 +72,17 @@ impl NJMat {
         )
     }
 
-    pub(super) fn compute_nj_q(&self) -> Mat {
+    pub(super) fn delta_tree_length(&self) -> DVector<f64> {
```
MattesMrzik: Even though `delta_tree_length` is a better name than `q` I would add a comment that this is the var commonly referred to as q in the literature. Bc other people who look at the code might not make that connection and by adding this comment we could help them understand.

phylo/src/tree/nj_matrices.rs
```diff
@@ -72,15 +72,17 @@ impl NJMat {
...
-            if r == c {
-                0.0
-            } else {
-                (n - 2) as f64 * self.distances[(r, c)] - s[r] - s[c]
+        let mut index = 0;
+        let mut delta_tree_len = DVector::zeros(n * (n + 1) / 2 - n);
+        for r in 1..n {
+            for c in 0..r {
+                delta_tree_len[index] = (n - 2) as f64 * self.distances[(r, c)] - s[r] - s[c];
+                index += 1
```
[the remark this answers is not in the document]
MattesMrzik: seems like a valid suggestion.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+    pub fn new(distance_function: D, rng: &'a R) -> Self {
+        info!("Creating regular NJTreeBuilder, first argmax choice for next pair of nodes to join");
+        Self {
+            randomise: Strategy::ArgMax,
+            distance_function,
+            rng,
+        }
+    }
+
+    /// Creates a Neighbor Joining Tree object with Softmax strategy, introduces stochasticity to tree building.
```
MattesMrzik: Nitpick: in the above comment, it's called `Neighbor Joining Tree Builder`, not just `Neighbor Joining Tree`.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+    /// Temperature can be between 0.0 and 1.0 and interpolates between the uniform and softmax distributions to select
+    /// the next pair of nodes to join. A temperature of 0.0 is fully uniform, while a temperature of 1.0 is fully softmax.
```
MattesMrzik, line 91: Perhaps also point out that the distances are fed to the softmax function.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+impl<D: EvolutionaryDistance, R: RandomSource> TreeBuilder for NJTreeBuilder<'_, D, R> {
+    fn build(&self, sequences: &Sequences) -> Result<Tree> {
+        let distances = self.compute_distance_matrix(sequences);
+        self.build_from_distances(distances, sequences)
+    }
+}
+
+impl<'a, D: EvolutionaryDistance, R: RandomSource> NJTreeBuilder<'a, D, R> {
+    /// Creates a Neighbor Joining Tree Builder object with Deterministic strategy, which uses argmin to minimize the tree length
+    pub fn new(distance_function: D, rng: &'a R) -> Self {
```
MattesMrzik: Why do we want to pass an `rng` to a deterministic strategy?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+                0.0, 0.0, 0.0, 0.2;
+                0.2, 0.2, 0.2, 0.0],
+        };
+        let sequences = Sequences::new(vec![
+            record!("A0", b""),
+            record!("B1", b""),
+            record!("C2", b""),
+            record!("D3", b""),
+        ]);
+        // Illegal for this test since it's DNA corrected distance
```
MattesMrzik: This comment confuses me. Do you mean that applying `LDNACorr` to AA is not correct?
Then I would say: "Applying LDNACorr to AA is not sound/appropriate, but here we are just using it for testing purposes."

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+        };
+        let sequences = Sequences::new(vec![
+            record!("A0", b""),
+            record!("B1", b""),
+            record!("C2", b""),
+            record!("D3", b""),
+        ]);
+
+        // FakeGen will return values that will select the last pair every time
+        let rng = FakeGen::from_u64_values(vec![5, 2, 0]);
```
MattesMrzik: Just to make sure I understand correctly: passing this rng to the builder will choose entries regardless of softmax values, right?
If so, does it actually test its implementation?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+            .build_from_distances(nj_distances, &sequences)
+            .unwrap();
+
+        // Uniform should be longer than the original NJ since it does not pick the optimal pair
+        assert!(nj_uniform_tree.length > nj_tree.length);
+        // Different resulting topologies
+        assert!(nj_uniform_tree.robinson_foulds(&nj_tree) > 0);
+
+        // Uniform should be longer than the original NJ since it does not pick the optimal pair
+        assert!(nj_softmax_tree.length > nj_tree.length);
```
MattesMrzik: This comment is confusing as it is not uniform but softmax

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+        // Different resulting topologies
+        assert!(nj_softmax_tree.robinson_foulds(&nj_tree) > 0);
+
+        // Since FakeGen always returns 0, both softmax trees should be the same
+        assert_eq!(nj_uniform_tree.nodes, nj_softmax_tree.nodes);
```
MattesMrzik: If they return the same tree, then the checks above for unifrom and softwax are checking the same tree against the normal nj tree. So these checks are kinda pointless then.

MattesMrzik (changes requested):
Looking good and locking forward to using this functionality. Thanks:) Made some small requests.

Also, perhaps add a test that runs the softmax tree creation twice and checking if different trees are returned, just as a sanity check? Although, the fake generator makes this kind of pointless...

phylo/src/tree/nj_matrices.rs
junniest: Good point, done!

phylo/src/tree/tree_builder.rs
junniest: Done (:

phylo/src/tree/nj_builder.rs
junniest: Not the distances, but the deltas. I will clarify in the docstrings!

phylo/src/tree/nj_builder.rs
junniest: Good q. Because the original NJ algorithm resolves ties by uniformly sampling among all nodes, but the current implementation simply picks the first value and doesn't use the rng. The tricky thing is that I could not force the FakeRandom to pick properly among several equally likely values and that messed with other tests, especially with TopologyOptimiser. I will add a todo there to get back to it once I refactor the randoms.
I have also now renamed Deterministic to ArgMax, since that's a much better description. In the docstring it says that the first value is always selected for now.

phylo/src/tree/nj_builder.rs
junniest: Yes, currently the rng will just spit out the index without any regard for what the distribution is. I will try to change that when I refactor the random module.
FakeGenerator did have some tests for it, and I've added more, just didn't want to commit here since it's different functionality. But since you brought it up I'll add it to this PR (:

junniest: Ah, actually, the FakeGenerator tests are already there, I added more for DefaultGenerator. Will still commit them here.

phylo/src/tree/nj_builder.rs
junniest: Yes, good point (: I think it's because I kept coming up with things to check, will remove.

phylo/src/tree/nj_builder.rs
junniest: Yes, you understood that correctly. Fixed the comment, especially because the distance function is only used in the definition of the struct since the distance matrix is provided for the test.
Actually, now that I think about it, the distance function is never used in the actual tests, we always pass a matrix. I think I should add a test that would use it.

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -143,7 +149,9 @@ impl<A: Alignment, AA: AncestralAlignment> PhyloInfoBuilder<A, AA> {
             Some(tree_file) => self.read_tree(tree_file)?,
             None => {
                 info!("Building NJ tree from sequences");
-                build_nj_tree_w_rng(&sequences, rng)?
+                self.tree_builder
+                    .unwrap_or(Box::new(NJTreeBuilder::new(LevenshteinDNACorrected, rng)))
+                    .build(&sequences)?
```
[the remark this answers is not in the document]
kalxed: I made this change locally if we want to add just for ease of switching the default tree builder later? Or maybe that is something we can configure globally?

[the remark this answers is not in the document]
junniest: That's not a big deal, I will bring it out, the issue is that having a "default" configuration isn't easily done at the moment because the "default" distance measure depends on the type of sequences that are read in. So in fact the current implementation isn't correct, because it will use the DNA distance regardless of the sequence type. I will fix this now, actually, so it's great that it was brought to my attention.

Right now the checks whether the sequence type and model type match are done dynamically, because previously I tried to do static checks for this and the typing quickly became too rigid to manage properly. I might have a better idea on how to manage this now by introducing a trait similar to what Mattes used for tree move compatibility with the specific cost functions (`optimisers/topo_optimiser/Compatible`).

The other thing is that I used to have a tendency to try and configure things globally, but this is not something that should be happening in the library code. The developers using the code (including myself) should set up the configuration to match the needs of their specific tool. The library should not have a persistent state in my opinion, as it is "just" a collection of tools/methods, and users/devs should take care of the overall configuration on their end.

junniest:
The rust version was bumped from 1.82.0 to 1.84.0 because the `isqrt` method is only stable since 1.84.0.

phylo/src/random/fake_random.rs
```diff
@@ -5,6 +5,7 @@ use rand::{RngCore, SeedableRng};
 /// signed (isize, i64, i32, i16, i8) integer types.
 /// Can also be set up to produce specific f64 values (up to floating point precision).
 /// If no pre-configured values are provided, it will return 0 for all integer types and 0.0 for f64.
+/// To make shuffle return the same order every time, provide u64::MAX as the only pre-configured value.
```
MattesMrzik: You mean that if you provide u64:MAX then the list will not get shuffled, right? Even if you pass no pre-configured values to the fake_rng the list will always get shuffled the same, right?

phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -24,31 +24,21 @@ use crate::{record_wo_desc as record, tree};
...
-
-                if let Ok(precomputed) =
-                    PIB::with_attrs(seq_file, tree_file).build()
+                let start_info = PIB::new(seq_file).build_w_rng(rng).unwrap();
+                let initial_cost = $builder::new(model.clone(), start_info).build().unwrap();
+cfg_if::cfg_if! {
+if #[cfg(not(feature="precomputed-test-results"))] {
+                TopologyOptimiser::new(initial_cost, $move_optimiser {}, rng).run().unwrap()
+} else {
+if let Ok(precomputed) =
```
MattesMrzik:
```suggestion
                if let Ok(precomputed) =
```

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -176,11 +182,9 @@ impl<A: Alignment, AA: AncestralAlignment> PhyloInfoBuilder<A, AA> {
         let sequences = self.read_sequences()?;
         let mut tree = match &self.tree_file {
             Some(tree_file) => self.read_tree(tree_file)?,
-            None => {
-                info!("Building NJ tree from sequences");
-                build_nj_tree_w_rng(&sequences, rng)?
-            }
+            None => self.build_nj_tree(rng, &sequences)?,
```
MattesMrzik: Could you add `info!` to these here too, like in the `build_w_rng`?

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,750 @@
...
+        // Interpolated probabilities, temp = 0.0 is the uniform distribution, temp = 1.0 is the softmax of distances
+        // Avoid copying the matrix by mutating in place
+        for element in exp_mat.iter_mut() {
+            *element = ((1.0 - temperature) * uniform_weight) + ((temperature) * *element);
+        }
+        exp_mat
+    }
+
+    /// Computes a one-hot vector indicating the position(s) of the minimum value(s) in the input vector.
+    /// Used to uniformly select among the minimum delta tree lengths.
```
MattesMrzik: Dont know if `one-hot` is an appropriate term here, since multiple indices can be 1. Usually with one-hot I expect a nested vector which with dimension [n_samples, n_classes].
Perhaps sth like this instead?
```suggestion
    /// Computes a vector indicating the position(s) of the minimum value(s) in the input vector,
    /// i.e., a vector with 1 at indices of the minimum, and 0 elsewhere.
    /// Used to uniformly select among the minimum delta tree lengths.
```

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,750 @@
...
+            ]
+        );
+        assert_eq!(softmax_vector.sum(), 1.0);
+    }
+
+    #[test]
+    fn softmax_examples() {
+        // Example values from https://medium.com/@hunter-j-phillips/a-simple-introduction-to-softmax-287712d69bac
+        let softmax = NJTreeBuilder::<LDNACorr>::softmax_from_deltas(dvector![-5.0, -7.0, -10.0]);
+        assert_relative_eq!(softmax, dvector![0.006, 0.047, 0.946], epsilon = 1e-3);
```
MattesMrzik, line 593: isnt the `softmax` test of the same form as the `softmax_examples?`. Then perhaps merge them into a single test.

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,750 @@
...
+        let softmax_vector = NJTreeBuilder::<LDNACorr>::softmax(delta_tree_length, 0.0);
+        assert_eq!(softmax_vector, dvector![0.2, 0.2, 0.2, 0.2, 0.2]);
+        assert_eq!(softmax_vector.sum(), 1.0);
+    }
+
+    #[test]
+    fn softmax_w_temp_between() {
+        let delta_tree_length = dvector![-1.3, -5.1, -2.2, -0.7, -1.1];
+        let softmax_vector = NJTreeBuilder::<LDNACorr>::softmax(delta_tree_length, 0.5);
+        assert_eq!(softmax_vector.sum(), 1.0);
```
MattesMrzik: to check interpolation, we could also do sth like this: 
```rust
let delta_tree_length = dvector![0.0, -1.0, 0.0, 0.0, 0.0];
let softmax_vector = NJTreeBuilder::<LDNACorr>::softmax(delta_tree_length, 0.5);
assert_eq!(softmax_vector, dvector![0.1, 0.6, 0.1, 0.1, 0.1]);
```

phylo/src/tree/nj_builder.rs
```diff
@@ -0,0 +1,750 @@
...
+            record!("D3", b""),
+        ]);
+
+        // FakeRng will return values that will select the last pair every time
+        let nj_softmax_tree = NJTreeBuilder::new_with_softmax(LDNACorr {}, 1.0)
+            .build_from_distances(
+                nj_distances.clone(),
+                &sequences,
+                &mut RandomGenerator::from_rng(FakeRng::from_u64_values(vec![5, 2, 0])),
+            )
```
MattesMrzik: Just out of curiosity: did you fiddle with the values in `FakeRng::from_u64_values(vec![5, 2, 0])` until they selected the last pair every time, or is there a smarter approach rather than just trial and error?

MattesMrzik (approved):
Looking good

phylo/src/random/fake_random.rs
junniest: The value sampled when the shuffle happens has to be `u64::MAX`. If no pre-configured values are provided then the default return value will be 0 and the list will get "shuffled" from e.g. 1, 2, 3, 4, 5 to 5, 1, 2, 3, 4.

In the previous version I was implementing shuffle manually so I could just skip the shuffling. Now because `FakeRng` implements `RngCore` shuffle is derived automatically from sampled u64 values. So I have no control over the exact way shuffle is done and cannot disable it.

In essence, almost every call any Rng methods goes through `next_u64()`, which returns the next u64. That's why I need the weird transformation of f64 to u64 and back. So, for example, if we sample 3 u64 values, 2 f64 values and then shuffle a list, we will in fact sample 6 u64 values. If we know the shuffle is last, we have to make sure that the 6th value in the pre-configured value list is `u64::MAX`. 

I hope this makes sense. I have no idea how to express this properly in the comment though, would gladly receive any suggestions from your side!

phylo/src/phylo_info/phyloinfo_builder.rs
junniest: Refactored building the tree into a method, since there's a copied piece of code in `build_with_ancestors_w_rng` as well.

phylo/src/random/fake_random.rs
MattesMrzik: Yes, this makes sense. For the comment, perhaps sth like this: 
```
/// Pre-configured values deterministically influece the order if shuffle is called.
/// If a no-shuffle is desired, provide u64::MAX as pre-configured value.
```

phylo/src/tree/nj_builder.rs
junniest: Agreed, I think I went with the goddamn Copilot suggestion and didn't check the term's appropriateness here.

phylo/src/random/fake_random.rs
junniest: Thanks, will use your suggestion!

phylo/src/tree/nj_builder.rs
junniest: Agreed, merged the two!

phylo/src/tree/nj_builder.rs
junniest: Yep, that makes sense. I updated the test to actually check that the mix works properly, this is what it looks like now:
```
   fn softmax_w_temp_between(#[case] temp: f64) {
        let delta_tree_length = dvector![-1.3, -5.1, -2.2, -0.7, -1.1];
        let softmax_temp = NJTreeBuilder::<LDNACorr>::softmax(delta_tree_length.clone(), temp);

        let softmax = NJTreeBuilder::<LDNACorr>::softmax(delta_tree_length.clone(), 1.0);
        let softmax_uniform = NJTreeBuilder::<LDNACorr>::softmax(delta_tree_length.clone(), 0.0);

        assert_relative_eq!(softmax_temp.sum(), 1.0);
        assert_eq!(
            softmax_temp,
            (softmax_uniform.scale(1.0 - temp) + softmax.scale(temp))
        );
    }
```

phylo/src/tree/nj_builder.rs
junniest: Actually, this is still from the previous impl of `FakeGenerator::sample` when it would return the indices, so this should not technically even be correct but it still is because the test tree is extremely simple. The way that sampling currently works is that I should actually provide the f64 values instead. Because the test tree is so simple and the best nodes to merge are always either the first or the last pair, now I create the `FakeRng` like this `FakeRng::from_f64_values(vec![1.0]))` both for regular NJ and softmax and it produces the same result.

PR #64, opened 2025-07-01
generic move_optimiser for spr and nni moves

### LINKED ISSUE #31: Move spr moves into separate class
Move the spr move functionality into a `TreeMove` class out of the `Tree` itself.

### LINKED ISSUE #32: Implement NNI moves
Either use the SPR move class of #31, as every NNI move is also an SPR move. Or implement it without using the SPR class. Whatever is simpler.

MattesMrzik:
Closes #31 and closes #32:
- `regraft_optimiser` is now reworked to be a generic `move_optimiser`
- both Spr and Nni implement this trait.


phylo/src/optimisers/spr_optimiser.rs
```diff
@@ -0,0 +1,311 @@
...
+use crate::likelihood::TreeSearchCost;
+use crate::optimisers::tree_mover::{MoveCostInfo, TreeMover};
+use crate::optimisers::BranchOptimiser;
+use crate::tree::{NodeIdx, Tree};
+use crate::Result;
+
+#[derive(Clone)]
+pub struct SprOptimiser {}
+
+impl TreeMover for SprOptimiser {
```
junniest: I'd suggest using `SprMover` and `NniMover` rather than `SprOptimiser` as here, just because otherwise it seems a bit confusing to me... I wasn't sure if there's some additional trait for Optimisers or is this a tree move. I see what you meant, because it's looking for the best move, but I think we should try to keep trait/struct names matching. 
Or, potentially, call the trait `MoveOptimiser`?..

phylo/src/optimisers/spr_optimiser.rs
```diff
@@ -0,0 +1,311 @@
...
+        self.find_max_cost_regraft_for_prune(base_cost, cost, node_idx)
+    }
+
+    fn move_locations<'a>(&self, tree: &'a Tree) -> impl Iterator<Item = &'a NodeIdx> {
+        tree.preorder().iter().filter(|&n| *n != tree.root)
+    }
+}
+
+impl SprOptimiser {
+    pub fn available_regraft_locations<'a>(
```
junniest: Did this need to be `pub`? I guess it has to because of the benchmarks, right? If it's only because of the benchmarks, I can introduce a feature to gate the `pub` from general access. If so, could you make an issue for it maybe?

phylo/src/optimisers/spr_optimiser.rs
```diff
@@ -0,0 +1,311 @@
...
+        bail!("Cannot regraft to root node.");
+    }
+    if regraft.parent == prune.parent {
+        bail!("Prune and regraft nodes must have different parents.");
+    }
+
+    Ok(rooted_spr_unchecked(tree, prune_idx, regraft_idx))
+}
+
+// TODO: alternatively to making this pub(crate) the tests can be moved to the bottom of this
```
junniest: I'm assuming this is your comment, Mattes, I agree (:

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -76,9 +89,10 @@ impl<C: TreeSearchCost + Clone + Display + Send> TopologyOptimiser<C> {
         }
         }
 
+        let tree_mover = self.c.tree_mover().clone();
         // The best move on this iteration might still be worse than the current tree, in which case
         // the search stops.
-        // This means that curr_cost is always hugher than or equel to prev_cost.
+        // This means that curr_cost is always higher than or equal to prev_cost.
```
junniest, line 152: nice catch (;

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -66,7 +74,12 @@ impl<C: TreeSearchCost + Clone + Display + Send> TopologyOptimiser<C> {
         let mut prev_cost = f64::NEG_INFINITY;
         let mut iterations = 0;
 
-        let possible_prunes: Vec<_> = init_tree.find_possible_prune_locations().copied().collect();
+        let possible_prunes: Vec<_> = self
```
junniest: I guess since we're in both spr and nni move territory now this var should be something like `possible_move_locations`? Or simply `move_locations`, since this would be too long?..

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::Result;
+use anyhow::bail;
+
+use crate::optimisers::{MoveCostInfo, TreeMover};
+use crate::tree::{
+    NodeIdx::{self, Leaf},
+    Tree,
+};
+
+use super::BranchOptimiser;
```
junniest: For some reason I don't really like using `super` uses. Not sure what the recommendations are in the Rust guidelines, I will check what they suggest.
In general, at the moment the use statements in the crate are sorted as follows:
```
// All things std
use std::f64;
use std::fmt::Display;

// All things from external crates
use anyhow::bail;

// All things from the phylo crate
use crate::likelihood::TreeSearchCost;
use crate::optimisers::{BranchOptimiser, MoveCostInfo, TreeMover};
use crate::tree::{NodeIdx::*, Tree,};
use crate::Result;
```
This way the uses are also sorted alpabetically.

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,143 @@
...
+    node_idx: &NodeIdx,
+    child_idx: &NodeIdx,
+    base_cost: f64,
+    mut cost_fn: C,
+) -> Result<MoveCostInfo> {
+    let mut new_tree = rooted_nni(cost_fn.tree(), node_idx, child_idx)?;
+    cost_fn.update_tree(new_tree.clone(), &[*node_idx]);
+    let mut move_cost = cost_fn.cost();
+    // TODO: do we really want to do run is only when its worse?
+    // if that is not the case. do we run this somewhere is (if the move is good in itself (even
```
junniest: Yes, the branch length optimisation runs after a move is selected and applied in the `topo_optimiser`, line 110. For the SPR moves only the branch length at the regraft point is optimised per attempted move.

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,143 @@
...
+        node_idx: &crate::tree::NodeIdx,
+    ) -> Result<Option<MoveCostInfo>>
+    where
+        C: TreeSearchCost<Self> + std::fmt::Display + Send + Clone + std::fmt::Display,
+    {
+        let mut max_cost_info = None;
+        let mut max_cost = f64::MIN;
+        println!("running nnis at {}, base_cost = {}", node_idx, base_cost);
+        for child_idx in &cost.tree().node(node_idx).children {
+            // TODO: is this cost.clone() really the way to go?
```
junniest: I think in the spr moves the costs are cloned to make parallelisation possible. I am not sure if cloning is necessary here and/or whether similar parallelisation makes sense since there's only 3 possible nni switches at any time.

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,143 @@
...
+        base_cost: f64,
+        cost: &C,
+        node_idx: &crate::tree::NodeIdx,
+    ) -> Result<Option<MoveCostInfo>>
+    where
+        C: TreeSearchCost<Self> + std::fmt::Display + Send + Clone + std::fmt::Display,
+    {
+        let mut max_cost_info = None;
+        let mut max_cost = f64::MIN;
+        println!("running nnis at {}, base_cost = {}", node_idx, base_cost);
```
junniest: Please use `info!` or `debug!` instead of `println!` so that everything goes in the log file.

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl TreeMover for NniOptimiser {
+    fn tree_move_at_location<C>(
+        &self,
+        base_cost: f64,
+        cost: &C,
+        node_idx: &crate::tree::NodeIdx,
+    ) -> Result<Option<MoveCostInfo>>
+    where
+        C: TreeSearchCost<Self> + std::fmt::Display + Send + Clone + std::fmt::Display,
```
junniest: `std::fmt::Display` appears twice here.
Also you already have the `use std::fmt::Display` statement, you can simply write `Display` here, I find it reads better.

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,143 @@
...
+
+pub(crate) fn rooted_nni(tree: &Tree, node_idx: &NodeIdx, child_idx: &NodeIdx) -> Result<Tree> {
+    if node_idx == &tree.root {
+        bail!("For the rooted NNI the node mustn't be the root of the tree.");
+    }
+    if matches!(node_idx, Leaf(_)) {
+        bail!("For the rooted NNI the node mustn't be a leaf");
+    }
+    if tree.node(child_idx).parent.is_none() || tree.node(child_idx).parent.unwrap() != *node_idx {
+        bail!("The provided child is not the child");
```
junniest: I am not sure I understand what you were trying to say with this error message (:

phylo/src/optimisers/nni_optimiser_tests.rs
```diff
@@ -0,0 +1,104 @@
+use itertools::Itertools;
+
+use crate::tree;
+use crate::tree::{NodeIdx::Leaf, Tree};
+
+use super::rooted_nni;
+
+#[cfg(test)]
+fn compare_trees(tree: &Tree, true_tree: Tree) {
```
junniest: Maybe this should actually be a part of the `Eq` trait implementation for the `Tree`?

phylo/src/optimisers/nni_optimiser_tests.rs
```diff
@@ -0,0 +1,104 @@
...
+    assert_eq!(tree.root, true_tree.root);
+    for node_idx in tree.preorder() {
+        let current = tree.node(node_idx);
+        let current_id = current.id.clone();
```
junniest: hm, is this clone necessary if you're using a reference to read it anyway?

phylo/src/optimisers/spr_optimiser.rs
```diff
@@ -0,0 +1,311 @@
+use anyhow::bail;
+use approx::relative_eq;
```
junniest: nitpick: Move these two down to sit with the `use itertools` and `use log`

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -118,62 +134,61 @@ impl<C: TreeSearchCost + Clone + Display + Send> TopologyOptimiser<C> {
     }
 }
 
+// TODO: why do we have a separate mod here?
```
junniest: I think Merlin moved it out at some point. Not sure it's necessary, especially since it's not just spr moves anymore.

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -118,62 +134,61 @@ impl<C: TreeSearchCost + Clone + Display + Send> TopologyOptimiser<C> {
...
 
+                dirty_nodes.push(*move_location);
                 if best_cost > base_cost {
-                    cost_fn.update_tree(best_tree, &[*prune, best_regraft]);
-                    info!(
-                        "    Regrafted to {:?}, new cost {}.",
-                        best_regraft, best_cost
-                    );
+                    cost_fn.update_tree(best_tree, &dirty_nodes);
+                    info!("    Moved tree, new cost {}.", best_cost);
```
junniest: I think this log message is confusing, it's not a tree that moved, it's that a tree was modified through a move. I think it would be wise to add a `Display` impl to the `tree_mover` and have it display the move type.
Also, I've started removing periods from ends of log messages because they look a bit silly in the whole log and I didn't use them consistently at the start.
So the log message could be something like this:
`info!("{tree_mover} move applied, new cost {best_cost}");`

phylo/src/substitution_models/mod.rs
```diff
@@ -134,28 +135,49 @@ impl<Q: QMatrix> SubstitutionCostBuilder<Q> {
             model: self.model,
             info: self.info,
             tmp,
+            tree_mover: SprOptimiser {},
+        })
+    }
+
+    pub fn build_with_nni(self) -> Result<SubstitutionCost<Q, NniOptimiser>> {
```
junniest: I think that rather than having an implicit "default" build method that uses SPRs and an additional one for an NNI mover it's best to leave the `build()` function as is, but rather make a new constructor and add the mover as a field in the builder.
E.g. the `new()` will by default construct the cost with SPRs, but `with_nni` will construct the cost with NNIs. This will keep the definition of the cost separate from the process of building it.

phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -267,9 +267,13 @@ fn pip_vs_subst_dna_tree() {
         .build()
         .unwrap();
     let k80 = SubstModel::<K80>::new(&[], &[4.0]);
-    let k80_res = TopologyOptimiser::new(SCB::new(k80.clone(), info.clone()).build().unwrap())
-        .run()
-        .unwrap();
+    let k80_res = TopologyOptimiser::new(
+        SCB::new(k80.clone(), info.clone())
+            .build_with_nni()
```
junniest: I am not a fan of this method, see my comment below.

phylo/src/optimisers/mod.rs
```diff
@@ -17,7 +23,8 @@ pub struct SingleValOptResult {
     pub value: f64,
 }
 
-pub struct PhyloOptimisationResult<C: TreeSearchCost> {
+pub struct PhyloOptimisationResult<C: TreeSearchCost<TM>, TM: TreeMover> {
```
junniest: This looks a little weird, feels like we should simplify this, but it's not clear yet to me how.

phylo/src/optimisers/tree_mover.rs
```diff
@@ -0,0 +1,46 @@
...
+
+pub trait TreeMover: Clone {
+    fn tree_move_at_location<C: TreeSearchCost<Self> + Display + Send + Clone + Display>(
+        &self,
+        base_cost: f64,
+        cost: &C,
+        node_idx: &NodeIdx,
+    ) -> Result<Option<MoveCostInfo>>;
+
+    // or should we also use the cost here such that it might be more likely that this method and
```
junniest: That seems like a good idea. The cost function contains the tree and it is passed to the `tree_move_at_location` so I am not sure why we need to pass the tree again.

phylo/src/pip_model/mod.rs
```diff
@@ -274,9 +277,18 @@ impl<Q: QMatrix> TreeSearchCost for PIPCost<Q> {
     fn tree(&self) -> &Tree {
         &self.info.tree
     }
+    // to avoid having the treemover as field of the cost
+    // i can just return a new instance here
+    // like below.
```
junniest: Sorry, I don't understand what you mean by below here.

phylo/src/tree/mod.rs
```diff
@@ -69,6 +68,8 @@ pub struct Tree {
     leaf_ids: Vec<String>,
     pub complete: bool,
     pub n: usize,
+    /// TODO: this is the total sum of blens and not the height of the tree
```
junniest: It's actually length, not height, my bad! Well spotted!

phylo/src/substitution_models/mod.rs
junniest: Actually, now that I'm thinking about it, should the TreeMover be part of the Cost function at all? I don't think so, the cost is just the likelihood, the type of move used for optimisation here is irrelevant.
It seems like it should be the other way around maybe, the TreeMover should incorporate the cost function and then only NNI moves are defined for TKF92?

phylo/src/substitution_models/mod.rs
```diff
@@ -176,9 +198,13 @@ impl<Q: QMatrix> TreeSearchCost for SubstitutionCost<Q> {
...
         &self.info.tree
     }
+
+    fn tree_mover(&self) -> &TM {
+        &self.tree_mover
+    }
 }
 
-impl<Q: QMatrix> ModelSearchCost for SubstitutionCost<Q> {
+impl<Q: QMatrix, TM: TreeMover> ModelSearchCost for SubstitutionCost<Q, TM> {
```
junniest: The comment above would also remove the need to inform the `ModelSearchCost` of the `TreeMover`, since it is irrelevant for model search.

phylo/src/pip_model/mod.rs
junniest: Did you mean like in the `phylo/src/substitution_models/mod.rs` file? Then the comments there apply here too.

junniest (commented):
I think there's a bit of logical inconsistency now where the cost functions all rely on having a tree mover attached, even if the cost function is only meant for model optimisation. I have a feeling that it's rather the `TreeMover` should have a `TreeSearchCost` attached than the other way around.

MattesMrzik:
I thought about alternative ways to implement this. One possibility could be to: have the `TopoOpti` to take a generic `TreeMover` and a generic `TreeSearchCost` and have a trait bound that forces them to be compatible, like so:
```rust
struct TopoOpti<C, M>
where
    C: TreeSearchCost + Compatible<M>,
    M: TreeMover,
{
    cost: C,
    mover: M,
}
trait Compatible<M: TreeMover>: TreeSearchCost {}
impl Compatible<Spr> for PipCost {}
impl Compatible<Nni> for PipCost {}
impl Compatible<Nni> for TkfCost {}
```
Which doesn't create the circular dependency between `Cost` and `TreeMover` like in the current implementation. Additionally, users of the `phylo` crate may implement their own `Cost` and `TreeMover` and declare compatibilities. The only downside may be that they are also able to add the line `impl Compatible<Spr> for TkfCost {}`, causing runtime errors.

junniest:
> I thought about alternative ways to implement this. One possibility could be to: have the `TopoOpti` to take a generic `TreeMover` and a generic `TreeSearchCost` and have a trait bound that forces them to be compatible, like so:
> 
> ```rust
> struct TopoOpti<C, M>
> where
>     C: TreeSearchCost + Compatible<M>,
>     M: TreeMover,
> {
>     cost: C,
>     mover: M,
> }
> trait Compatible<M: TreeMover>: TreeSearchCost {}
> impl Compatible<Spr> for PipCost {}
> impl Compatible<Nni> for PipCost {}
> impl Compatible<Nni> for TkfCost {}
> ```

I like this! This seems like a nice way to restrict behaviour. We will need to document this very properly though! 

> Which doesn't create the circular dependency between `Cost` and `TreeMover` like in the current implementation. Additionally, users of the `phylo` crate may implement their own `Cost` and `TreeMover` and declare compatibilities. The only downside may be that they are also able to add the line `impl Compatible<Spr> for TkfCost {}`, causing runtime errors.

Actually, if we define the trait and the structs in `phylo`, people using it as an external crate will not be allowed to add this `impl Compatible<Spr> for TkfCost {}`. It's because of the orphan rule [implementing traits](https://doc.rust-lang.org/book/ch10-02-traits.html#implementing-a-trait-on-a-type), it says that "But we can’t implement external traits on external types". So that would just not be allowed by the compiler, but implementing `Compatible<Nni>` for some new type will be allowed.

So this could be the perfect way to handle that! I would maybe rename it to `CompatibleMove<Spr>`? Or some other name that would read properly in the code and would mean that "this cost function is compatible with this move specifically".

phylo/src/optimisers/blen_optimiser.rs
```diff
@@ -10,21 +11,25 @@ use crate::optimisers::{PhyloOptimisationResult, SingleValOptResult};
...
     pub fn new(cost: C) -> Self {
         Self {
+            phantom: PhantomData,
             epsilon: 1e-3,
             c: RefCell::new(cost),
         }
     }
 
-    pub fn run(mut self) -> Result<PhyloOptimisationResult<C>> {
+    pub fn run(mut self) -> Result<PhyloOptimisationResult<C, TM>> {
```
junniest: Hm, actually, the `BranchOptimiser` doesn't rely on the `TreeMover`, right? I don't think you should be the one dealing with this, but I am thinking I could potentially make the code simpler by adding a `BranchSearchCost` trait and removing a silly check from the `SprOptimiser`, line 199: 
```
if cost_fn.blen_optimisation() && move_cost <= base_cost {
```

I'll think about that.

phylo/src/optimisers/blen_optimiser.rs
```diff
@@ -72,7 +78,7 @@ impl<C: TreeSearchCost + Clone + Display> BranchOptimiser<C> {
     }
 }
 
-impl<C: TreeSearchCost + Clone + Display> BranchOptimiser<C> {
+impl<C: TreeSearchCost<TM> + Clone + Display, TM: TreeMover> BranchOptimiser<C, TM> {
```
junniest: Refactoring the `TreeSearchCost` to two traits, `TreeSearchCost` and `BranchSearchCost` will also remove the need for `Phantom`s here.

phylo/src/optimisers/mod.rs
junniest: I am guessing that your suggestion of the `Compatible` trait will remove the need to have a `Phantom` here as well, which would be great!

phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -28,7 +28,7 @@ macro_rules! define_optimise_trees {
                 seq_file: &std::path::Path,
                 _: &std::path::Path,
                 model: $model<Q>,
-            ) -> PhyloOptimisationResult<$cost<Q>> {
+            ) -> PhyloOptimisationResult<$cost<Q, SprOptimiser>, SprOptimiser> {
```
junniest: I guess we won't need this doubling with the compatibility trait added either?

junniest (commented):
I really like the `Compatible` trait idea! We will have to document it pretty thoroughly, but I think that's the way to go to make sure there's proper compile-time checks for this stuff.

phylo/src/pip_model/mod.rs
MattesMrzik: Due to a rework, this is not relevant anymore

phylo/src/optimisers/tree_mover.rs
MattesMrzik: `tree_move_at_location` must always get the most up to date cost, so we pass the cost there. But before we call this fn we need to calculate all possible branches at which we can make a move, and for that we also need the tree. So it might be clearer to pass the cost (which contains the tree) to both of the trait fns. 
>  so I am not sure why we need to pass the tree again.

To not pass it again: do you mean to set it as field of `TreeMover`? But we still need to update its tree after each move. So i think passing it makes it more clear that we need to consider the most up-to-date tree.

phylo/src/optimisers/topo_optimiser_tests.rs
MattesMrzik: Due to a rework, this is not relevant anymore

phylo/src/optimisers/topo_optimiser.rs
MattesMrzik: done

phylo/src/optimisers/topo_optimiser.rs
MattesMrzik: Yep:) (it is `move_locations` already)

phylo/src/optimisers/spr_optimiser.rs
MattesMrzik: I managed to remove the pub without affecting the benches

phylo/src/optimisers/spr_optimiser.rs
MattesMrzik: I like `MoveOptimiser`, too.

phylo/src/optimisers/topo_optimiser.rs
MattesMrzik: moved the `spr::fold_improving_moves` to `TopologyOptimiser::fold_improving_moves`

phylo/src/optimisers/nni_optimiser_tests.rs
MattesMrzik: Nope, it is not necessary, thanks for pointing that out.

phylo/src/optimisers/nni_optimiser.rs
MattesMrzik: is `"The provided child_idx (i.e. the node that indicates which subtrees should be swapped) is not a child of the provided node_idx."` better?
```
/// .           |
/// .      -- parent --
/// .      |          |
/// .  --node--      sibling
/// .  |      |
/// .  .    child
```

the parameter `child` indicates the subtree (either the left or right child of `node`) which should be swapped with `sibling`.

phylo/src/optimisers/nni_optimiser.rs
MattesMrzik: Ok. For now, I won't do any parallelization, but may investigate this in the future. Since for every move I have to do the dynamic programming reestimation of the two nodes, so it might be worth to parallelize.

phylo/src/optimisers/nni_optimiser.rs
MattesMrzik: Before merge i will make a commit that removes/replaced println, this was just for me to understand the code (when runnings tests)

phylo/src/optimisers/mod.rs
MattesMrzik: yep:)

phylo/src/optimisers/blen_optimiser.rs
MattesMrzik: For now, i won't split this into two traits. We may open another issue: ... (coming soon)

phylo/src/optimisers/blen_optimiser.rs
MattesMrzik: With the `Compatible` trait it no longer depends on the `TreeMover`. We may open another issue: ... (coming soon)

phylo/src/optimisers/nni_optimiser.rs
MattesMrzik: Ok, this is nice, I will try to not overlook any files.

phylo/src/optimisers/nni_optimiser_tests.rs
MattesMrzik: See #78

phylo/src/substitution_models/mod.rs
MattesMrzik: See the new approach with the trait `Compatible`

phylo/src/optimisers/move_optimiser.rs
```diff
@@ -0,0 +1,65 @@
...
+        MoveCostInfo {
+            cost,
+            dirty_nodes,
+            tree,
+        }
+    }
+
+    /// The nodes that where affected by the move operation. These are passed to the
+    /// [`TreeSearchCost::update_tree`], such that the cost may update its internal node states.
+    pub fn dirty_nodes(&self) -> &Vec<NodeIdx> {
```
junniest: nitpick: I think this can be &[NodeIdx], since the container itself doesn't matter.

phylo/src/optimisers/move_optimiser.rs
```diff
@@ -0,0 +1,65 @@
...
+/// This is the result of a move operation in a phylogenetic tree, i.e. the result of
+/// a call to [`MoveOptimiser::best_move_at_location`].
+pub struct MoveCostInfo {
+    cost: f64,
+    tree: Tree,
+    dirty_nodes: Vec<NodeIdx>,
+}
+
+impl MoveCostInfo {
+    pub(crate) fn new(cost: f64, tree: Tree, dirty_nodes: Vec<NodeIdx>) -> Self {
```
junniest: I wonder if `new` should also just be public? In case someone implements a different move time externally, they'd need to create this struct.

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,248 @@
...
+use crate::Result;
+
+#[derive(Clone)]
+pub struct NniOptimiser {}
+
+impl MoveOptimiser for NniOptimiser {
+    fn move_locations<'a, C: TreeSearchCost + Display + Send + Clone + Display>(
+        &self,
+        cost: &'a C,
+    ) -> impl Iterator<Item = &'a crate::tree::NodeIdx> {
```
junniest: nitpick: this specification (`crate::tree::NodeIdx`) is not necessary (here and in line 32).

phylo/src/optimisers/move_optimiser.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Calculates the best move at the given location in the tree.
+    /// E.g. for SPR, the given location `node_idx` is the prune location. Then, all
+    /// possible regraft locations are iterated over, and the one with the best cost is
+    /// returned.
+    fn best_move_at_location<C: TreeSearchCost + Display + Send + Clone + Display>(
+        &self,
+        base_cost: f64,
+        cost: &C,
+        node_idx: &NodeIdx,
+    ) -> Result<Option<MoveCostInfo>>;
```
junniest: I think we should remove the `Option`. Judging from what the NNI optimiser is doing, it will always return the best move, even if the move isn't better than the starting topology. Same for the SPR moves, the move should be returned no matter what, even if the resulting likelihood is lower than the original. This could allow us to potentially take suboptimal moves if necessary.

phylo/benches/spr.rs
```diff
@@ -45,18 +46,17 @@ fn run_single_spr_cycle_for_sizes<Q: QMatrix + QMatrixMaker + Send>(
             bench.iter_batched(
                 // clone because of interior mutability in PIPCost
                 || data.clone(),
-                |(cost_fn, prune_locations)| single_spr_cycle(cost_fn, prune_locations),
+                |(cost_fn, prune_locations)| {
+                    single_spr_cycle(cost_fn, prune_locations, SprOptimiser {})
```
junniest: Minor: I know that these are benches and not production code, but I have a feeling that passing a new instance of the `SprOptimiser` in every place where it is necessary might be problematic. For example, if we change the way we initialise it, we could end up with different configurations in different places. I suggest cloning the `regraft_optimiser` var instead of making a new instance everywhere.

I also think we should implement `new` and potentially `default` for the optimisers ([relevant api guideline](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits)). This kind of initialisation, I think, will not work in an external crate, especially if the `SprOptimiser` gains any additional functionality.

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,248 @@
...
+            move_cost = blen_opt.final_cost;
+            new_tree.set_blen(node_idx, blen_opt.value);
+        }
+    }
+    Ok(MoveCostInfo::new(move_cost, new_tree, vec![*node_idx]))
+}
+
+fn rooted_nni(tree: &Tree, node_idx: &NodeIdx, child_idx: &NodeIdx) -> Result<Tree> {
+    if node_idx == &tree.root {
+        bail!("For the rooted NNI the node mustn't be the root of the tree");
```
junniest, line 92: Question: wouldn't an NNI move work just fine on the root node? SPR moves are not allowed on the root because of the pruning, otherwise the root node will have a branch length without meaning, but in the case of NNIs the children of the root can be swapped around no problem?

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,248 @@
...
+    new_tree.compute_preorder();
+    new_tree.compute_postorder();
+    debug_assert_eq!(new_tree.postorder().len(), new_tree.preorder().len());
+    debug_assert_eq!(new_tree.postorder().len(), tree.postorder().len());
+    new_tree
+}
+
+#[cfg(test)]
+#[cfg_attr(coverage, coverage(off))]
+pub mod private_nni_tests {
```
junniest: nitpick: I had the `pub` qualifiers in private tests for some reason, but I am removing them now. You can ditch it immediately.

phylo/src/optimisers/move_optimiser.rs
junniest: I see that the SPR optimiser actually returns a `None` when then previous topology change made a prune location ineligible. I think that's an artefact of how I implemented it and it shouldn't be like that. I'll leave a suggestion for a workaround for the `SprOptimiser` because I still think that we should not have an `Option` here, it's too much nesting.

phylo/src/optimisers/spr_optimiser.rs
```diff
@@ -0,0 +1,432 @@
...
+    fn find_max_cost_regraft_for_prune<C: TreeSearchCost + Clone + Display + Send>(
+        &self,
+        base_cost: f64,
+        cost: &C,
+        prune_location: &NodeIdx,
+    ) -> Result<Option<MoveCostInfo>> {
+        let tree = cost.tree();
+        if tree.children(&tree.root).contains(prune_location) {
+            // due to topology change the current node may have become the direct child of root
+            return Ok(None);
```
junniest: I think here we can work around the need for `Option` by returning `Ok(MoveCostInfo::new(base_cost, tree, vec![]))`.

phylo/src/optimisers/spr_optimiser.rs
```diff
@@ -0,0 +1,432 @@
...
+        // reoptimise branch length at the regraft location
+        let mut o = BranchOptimiser::new(cost_fn);
+        let blen_opt = o.optimise_branch(&regraft)?;
+        if blen_opt.final_cost > move_cost {
+            move_cost = blen_opt.final_cost;
+            new_tree.set_blen(&regraft, blen_opt.value);
+        }
+    }
+    debug!("    Regraft to {regraft:?} w best cost {move_cost}");
+    Ok(MoveCostInfo::new(move_cost, new_tree, vec![regraft]))
```
junniest: This is new, and I think I missed this last time, but this should definitely be:
`Ok(MoveCostInfo::new(move_cost, new_tree, vec![prune_location, regraft]))`
For SPR moves the tree loses a chunk at the prune location so the intermediate values need to be recalculated from that point onwards, otherwise the likelihood will be incorrect.

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -68,15 +106,15 @@ impl<C: TreeSearchCost + Clone + Display + Send> TopologyOptimiser<C> {
...
+    /// use phylo::optimisers::{SprOptimiser, TopologyOptimiser};
     /// use phylo::phylo_info::PhyloInfoBuilder;
     /// use phylo::substitution_models::{SubstModel, SubstitutionCostBuilder, K80};
     ///
     /// let info = PhyloInfoBuilder::new("./examples/data/K80.fasta").build()?;
     /// let k80 = SubstModel::<K80>::new(&[], &[4.0, 1.0]);
     /// let c = SubstitutionCostBuilder::new(k80, info).build()?;
     /// let unopt_cost = c.cost();
-    /// let optimiser = TopologyOptimiser::new(c);
+    /// let optimiser = TopologyOptimiser::new(c, SprOptimiser {});
```
junniest, line 118: Yes, we definitely need a `new` for the optimisers.

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -144,64 +184,52 @@ impl<C: TreeSearchCost + Clone + Display + Send> TopologyOptimiser<C> {
...
         debug_assert!(
             {
-                let correct_prune_locations =
-                    cost_fn.tree().find_possible_prune_locations().collect_vec();
-                prune_locations
+                let correct_move_locations = move_opti.move_locations(cost_fn).collect_vec();
+                move_locations
                     .iter()
-                    .all(|prune_location| correct_prune_locations.contains(prune_location))
+                    .all(|prune_location| correct_move_locations.contains(prune_location))
```
junniest: nitpick: rename `prune_location` -> `move_location`

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -144,64 +184,52 @@ impl<C: TreeSearchCost + Clone + Display + Send> TopologyOptimiser<C> {
...
-    use log::info;
-
-    use crate::likelihood::TreeSearchCost;
-    use crate::optimisers::RegraftOptimiser;
-    use crate::tree::NodeIdx;
-    use crate::Result;
 
     /// Iterates over `prune_locations` in order and applies the best (improving)
-    /// SPR move for each pruneing location in place
+    /// tree move for each pruning location in place
```
junniest: nicpick: rename `prune_locations` -> `move_locations` and `pruning` -> `move` in the comment

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -144,64 +184,52 @@ impl<C: TreeSearchCost + Clone + Display + Send> TopologyOptimiser<C> {
...
-                    best_regraft_info.cost(),
-                    best_regraft_info.regraft(),
-                    best_regraft_info.into_tree(),
+                let (best_cost, mut dirty_nodes, best_tree) = (
+                    move_cost_info.cost(),
+                    move_cost_info.dirty_nodes().clone(),
+                    move_cost_info.into_tree(),
                 );
 
+                dirty_nodes.push(*move_location);
```
junniest: If the proper dirty nodes are returned by the `SprOptimiser` (both prune and regraft locations which are known there) there will be no need to mutate the `dirty_nodes`, which will avoid a clone and is in general always a good idea.

phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -227,13 +259,38 @@ fn k80_sim_data_vs_phyml_wrong_start() {
...
+    let res = optimise_tree_nni(&seq_file, &tree_file, wag.clone());
+    assert!(res.final_cost >= res.initial_cost);
+    let wag_tree = res.cost.tree();
+
+    let phyml_res = PIB::with_attrs(seq_file.clone(), fldr.join("phyml_nogap.newick"))
+        .build()
+        .unwrap();
+    let phyml_logl = SCB::new(wag, phyml_res.clone()).build().unwrap().cost();
+
+    // Compare tree height and logl to the output of PhyML
```
junniest, line 288: Oh, it still converges to the same result with NNIs? Nice!

phylo/src/parsimony/cost/basic_parsimony_cost.rs
```diff
@@ -27,7 +27,9 @@ impl BasicParsimonyCost {
         let tmp = RefCell::new(BasicParsimonyInfo::new(&info));
         Ok(BasicParsimonyCost { info, tmp })
     }
+}
 
+impl BasicParsimonyCost {
```
junniest: nitpick: I think this new `impl BasicParsimonyCost` appeared by accident.

phylo/src/parsimony/cost/dollo_parsimony_cost.rs
```diff
@@ -39,13 +39,13 @@ impl DolloParsimonyCost<SimpleScoring> {
...
 }
-
 impl<S: ParsimonyScoring> DolloParsimonyCost<S> {
     pub fn with_scoring(info: PhyloInfo, scoring: S) -> Self {
         let tmp = RefCell::new(DolloParsimonyInfo::new(&info));
         DolloParsimonyCost { info, tmp, scoring }
     }
-
+}
+impl<S: ParsimonyScoring> DolloParsimonyCost<S> {
```
junniest: nitpick: again, seems like this appeared by accident and isn't necessary.

phylo/src/substitution_models/mod.rs
```diff
@@ -155,7 +155,10 @@ impl<Q: QMatrix> Clone for SubstitutionCost<Q> {
     }
 }
 
-impl<Q: QMatrix> TreeSearchCost for SubstitutionCost<Q> {
+impl<Q> TreeSearchCost for SubstitutionCost<Q>
```
junniest: nitpick: why was this change necessary?

phylo/src/tree/mod.rs
```diff
@@ -68,8 +67,10 @@ pub struct Tree {
     preorder: Vec<NodeIdx>,
     leaf_ids: Vec<String>,
     pub complete: bool,
+    /// The number of leaves in the tree.
     pub n: usize,
-    pub height: f64,
+    /// The sum of all branch lengths of the tree.
+    pub magnitude: f64,
```
junniest: I totally agree that this is not the tree height, but I would prefer using `length` here. It is the total length of the tree really, not just length, but this is the term most commonly used for this value. `magnitude` is not a term I've ever seen used for this, and I think we shouldn't invent it just yet.

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    use crate::tree;
+    use crate::tree::Tree;
```
junniest, line 155: nitpick: combine these use statements into one.

junniest (commented):
It's looking great! Still some minor issues, but I think this is it (:

phylo/src/optimisers/move_optimiser.rs
MattesMrzik: Okey. Since in the topo optimiser `dirty_nodes.push(*move_location);` is done. I have to change `move_cost_info.dirty_nodes().clone()`, to `move_cost_info.dirty_nodes().to_vec().clone(),`

phylo/src/tree/mod.rs
MattesMrzik: Ok, renaming it to `length`

phylo/src/optimisers/nni_optimiser.rs
MattesMrzik: like so: `use crate::{tree, tree::Tree};` ?

phylo/src/substitution_models/mod.rs
MattesMrzik: this is just an artifact from the time when the cost would also depend on the treemover. I reverted it to the single line version.

phylo/src/optimisers/move_optimiser.rs
junniest: Hm, or maybe this should just return Vec<NodeIdx> and consume self? To eliminate cloning downstream...

phylo/src/optimisers/nni_optimiser.rs
MattesMrzik: the node defines the edge (node, pa(node)) that is used as the edge, connecting 4 subtrees, for which 2 of those subtrees are swapped. So if node = root then the edge (root, pa(root)) does not exist.

phylo/src/optimisers/spr_optimiser.rs
MattesMrzik: Ok, then we must (i think) pass `tree.clone()` to the new `MoveCostInfo`

phylo/src/optimisers/topo_optimiser.rs
MattesMrzik: Since we are calling `move_cost_info.into_tree()` which takes ownership of the move_cost_info we cannot keep a ref to the dirty nodes. so we must call to_vec(). (clone() does not work)

phylo/src/optimisers/topo_optimiser_tests.rs
MattesMrzik: yep:)

phylo/benches/spr.rs
MattesMrzik: Do you mean initializing `    let spr_optimiser = SprOptimiser {};`
before the loop?

Oke, we can add a new() to the trait (which currently takes no args, but may in the future, (do i understand this correctly?)), and the default() for the trait should never take any args, right?

phylo/src/optimisers/move_optimiser.rs
MattesMrzik: I removed the `Option` and implemented your suggestions.

phylo/src/optimisers/move_optimiser.rs
MattesMrzik: but we also need to consume it to get the tree from the `MoveCostInfo`

junniest: Yeah, so let's leave it like this for now, and I'll think about how to avoid cloning since we're consuming the object anyways. But could you add a TODO there so that I remember to think about it?

phylo/benches/spr.rs
junniest: Yes, exactly. I am not sure if this should be part of the trait or can just be an impl of SprOptimiser, it should probably be the latter. Things returning `Self` can be problematic in traits and I still don't entirely understand why. I'm sorry I'm like this, but the Rust guidelines (lol) say that all new types should implement both `new` and `default`, even if they are the exact same. I will wait until you merge and then add these wherever they are missing as well as implement the standard list of traits (`Copy`, `Clone`, `Eq`, `PartialEq`, `Ord`, `PartialOrd`,`Hash`,`Debug`, `Display`, `Default`).

phylo/src/optimisers/nni_optimiser.rs
junniest: Ah, fair enough, gotcha! Ignore this then.

phylo/src/optimisers/spr_optimiser.rs
junniest: Yes, you're right, it's used by reference here. That's not a big deal, since the spr optimiser clones a whole bunch of costs downstream, this isn't going to be a bottleneck anyways.

phylo/src/optimisers/topo_optimiser.rs
junniest: Fair! But at least it gets rid of the additional `mut`, which is a bit of a hidden side effect, which is good on its own.

phylo/src/optimisers/nni_optimiser.rs
junniest: Oh no, sorry, ignore this! I forgot that it's a crate-level macro, it should be the way that you had it.

phylo/src/optimisers/move_optimiser.rs
MattesMrzik: we could make the fields of `MoveCostInfo` public, then we can do 
```rust
let MoveCostInfo { cost, tree, dirty_nodes } = move_cost_info;
```
to avoid cloning

phylo/src/tree/mod.rs
MattesMrzik: Perhaps, we want another name, since `tree.len()` is the number of nodes and `tree.length`is the sum of all blens, and these names might be too similar.

junniest: I think we may want to rename `len()` instead. It's not a very meaningful name and it requires an `is_empty()` method as well, which isn't meaningful for this structure. We can rename `len()` to `taxa()` or `n()`? `leaves()` is already a method that returns a list of all leaves, so that's not a good idea. N is a common letter for taxa numbers, but could still be confusing.

I'd say you can ignore this for now and I will refactor the tree module once this is merged (and potentially Kai's changes).

phylo/src/optimisers/move_optimiser.rs
junniest: Or `pub(crate)`? A `new()` should be public, but I think we can get away with reading like this but not exposing the fields publicly.

phylo/src/optimisers/tree_mover.rs
junniest: I think I meant what your first paragraph implies. Passing the cost seems like a better idea than decoupling the tree from the cost and I think that this method only needs read access anyway.

phylo/src/optimisers/move_optimiser.rs
MattesMrzik: Sure, that would also work. Was just thinking that since `MoveCostInfo` is return by a trait fn we might as well make it pub

phylo/src/optimisers/nni_optimiser.rs
junniest: That's better, yes, but still a little confusing...
Since all the other error messages are in terms of the parent node, maybe something like:
`"The node {node_idx} must be the parent of {child_idx}"`.
I think that including the node indices will help understand the error.

phylo/src/optimisers/move_optimiser.rs
junniest: Ah, true. If it's an otherwise methodless struct (no complex internal logic as far as I can see) we can make it `pub`.

phylo/src/optimisers/tree_mover.rs
MattesMrzik: Am I understanding correctly, that i can resolve this comment now?

junniest: yes!

phylo/benches/spr.rs
```diff
@@ -40,23 +40,23 @@ fn run_single_spr_cycle_for_sizes<Q: QMatrix + QMatrixMaker + Send>(
...
                 || data.clone(),
-                |(cost_fn, prune_locations)| single_spr_cycle(cost_fn, prune_locations),
+                |(cost_fn, prune_locations)| {
+                    single_spr_cycle(cost_fn, prune_locations, spr_optimiser.clone())
+                },
                 criterion::BatchSize::SmallInput,
             );
         });
     };
+    let move_opti = SprOptimiser {};
```
junniest: minor: This should also be `spr_optimiser`

phylo/src/optimisers/nni_optimiser.rs
```diff
@@ -0,0 +1,259 @@
...
+#[cfg(test)]
+#[cfg_attr(coverage, coverage(off))]
+mod private_nni_tests {
+
+    use super::*;
+    use crate::tree;
+    use crate::tree::Tree;
+
+    #[cfg(test)]
+    fn compare_trees(tree: &Tree, true_tree: Tree) {
```
junniest, line 158: Minor: for a quick check you can also use the `robinson_foulds()` method of `Tree`. It will not check the branch lengths, but it will make sure that the topologies are the same, even when the root is different.

phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -227,13 +259,41 @@ fn k80_sim_data_vs_phyml_wrong_start() {
...
+    assert_relative_eq!(tree.length, phyml_info.tree.length, epsilon = 1e-4);
     assert_eq!(tree.robinson_foulds(&phyml_info.tree), 0);
 }
 
 #[test]
 #[cfg_attr(feature = "ci_coverage", ignore)]
-fn wag_no_gaps_vs_phyml_nj_tree_start() {
+fn wag_no_gaps_vs_phyml_nj_tree_start_nni() {
+    // TODO: this test only passes because the NNI moves dont run into a local optimum and therefore
+    // find the same best tree as the SprOptimiser. So, if the data chances the test may fail.
```
junniest: nitpick (typo): `chances` -> `changes`

junniest (approved):
Awesome stuff!

phylo/src/tree/mod.rs
MattesMrzik: See #82

phylo/benches/spr.rs
MattesMrzik: See #83.

phylo/src/optimisers/blen_optimiser.rs
MattesMrzik: See #84.

phylo/src/optimisers/blen_optimiser.rs
MattesMrzik: See #84.

PR #86, opened 2025-08-05
Release prep/macro use everywhere

junniest:
Using `record!` macro wherever `bio::io::fasta::Record` is being created + a warning for the `tree!` macro because it does not check newick validity.

### DIFF
--- a/phylo/src/alphabets/tests.rs
+++ b/phylo/src/alphabets/tests.rs
@@ -1,10 +1,10 @@
-use bio::io::fasta::Record;
 use rstest::*;
 
 use crate::alphabets::{
     dna_alphabet, protein_alphabet, ParsimonySet, AMB_AMINOACIDS, AMB_NUCLEOTIDES, AMINOACIDS,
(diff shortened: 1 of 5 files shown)


MattesMrzik (approved):

phylo/src/alphabets/tests.rs
kalxed, line 1: This seems good to me since it's a test file, otherwise I would advise against using alias for record for sake of confusion.

phylo/src/macros.rs
kalxed, line 1: Should this be an issue to fix? Or is declaring it fine? Also could be a debug log message just in case? Though I'm not entirely sure what the debug log on from_newick might be otherwise.

kalxed (approved):

phylo/src/macros.rs
junniest: The newick parser will crash with a semi-meaningful error (: I think it's ok like that for now, it was made for convenience, that's why there are no checks.

PR #88, opened 2025-08-11
Feature/random provider

junniest:
Implemented `RandomSource` to use in the crate and added a fake random to use in testing.

### DIFF
--- a/phylo/examples/rng_examples.rs
+++ b/phylo/examples/rng_examples.rs
@@ -1,50 +1,40 @@
 use rand::rngs::{SmallRng, StdRng};
 
-use phylo::random::{
-    gen, gen_probability, gen_range, global_rng, init_rng, GlobalRng, RandomGenerator,
-};
+use phylo::random::{DefaultGenerator, RandomGenerator, RandomSource};
 
 fn main() {
-    // 1. Using the global RNG (StdRng-based)
+    // 1. Using the default RNG (StdRng-based)
     let seed = 42;
-    println!("1. Global RNG (StdRng) with seed {seed} - reproducible:");
-    init_rng(seed);
+    println!("1. RNG (StdRng) with seed {seed} - reproducible:");
+    let rng = DefaultGenerator::new(seed);
     for i in 0..3 {
-        println!("  {i}: {:.6}", gen::<f64>());
+        println!("  {i}: {:.6}", rng.gen::<f64>());
     }
     // Reset and show reproducibility
-    init_rng(seed);
+    rng.reseed(seed);
     println!("After resetting to seed {seed}:");
     for i in 0..3 {
-        println!("  {i}: {:.6}", gen::<f64>());
+        println!("  {i}: {:.6}", rng.gen::<f64>());
     }
     println!();
 
-    // 2. Generating from global RNG:
-    println!("2. Generating from global RNG:");
-    println!("  Random f64: {:.6}", global_rng().gen::<f64>());
-    println!("  Random u32: {}", global_rng().gen::<u32>());
-    println!("  Random i32: {}", global_rng().gen::<i32>());
-    println!("  Probability [0,1): {:.6}", global_rng().gen_f64());
-    println!("  Range 1-100: {}", global_rng().gen_range(1..=100));
-    println!("  Range 0.0-10.0: {:.3}", global_rng().gen_range(0.0..10.0));
-    println!("  Boolean (50%): {}", global_rng().gen_bool(0.5));
-    println!("  Boolean (25%): {}", global_rng().gen_bool(0.25));
+    // 2. Generating from default RNG:
+    println!("2. Generating from default RNG:");
+    let rng = DefaultGenerator::default();
+    println!("  Random f64: {:.6}", rng.gen::<f64>());
+    println!("  Random u32: {}", rng.gen::<u32>());
+    println!("  Random i32: {}", rng.gen::<i32>());
+    println!("  Probability [0,1): {:.6}", rng.gen::<f64>());
+    println!("  Range 1-100: {}", rng.gen_range(1..=100));
+    println!("  Range 0.0-10.0: {:.3}", rng.gen_range(0.0..10.0));
+    println!("  Boolean (50%): {}", rng.gen_bool(0.5));
+    println!("  Boolean (25%): {}", rng.gen_bool(0.25));
     println!();
 
-    // 3. Generating using global RNG functions
-    println!("3. Generating using global RNG functions:");
-    println!("  Probability [0,1): {:.6}", gen_probability());
-    println!("  Random u32: {}", gen::<u32>());
-    println!("  Random i32: {}", gen::<i32>());
-    println!("  Range 1-10: {}", gen_range(1..=10));
-    println!("  Range -5.0 to 5.0: {:.3}", gen_range(-5.0..5.0));
-    println!();
-
-    // 4. Creating a custom StdRng instance
+    // 3. Creating a custom StdRng instance
     let seed = 123;
-    println!("4. Custom StdRng instance with seed {seed} - reproducible:");
-    let custom_std_rng: GlobalRng = RandomGenerator::new(seed);
+    println!("3. Custom StdRng instance with seed {seed} - reproducible:");
+    let custom_std_rng = RandomGenerator::<StdRng>::new(seed);
     for i in 0..3 {
         println!("  {i}: {:.6}", custom_std_rng.gen::<f64>());
     }
@@ -56,11 +46,11 @@ fn main() {
     }
     println!();
 
-    // 5. Generating from the custom StdRng instance
-    println!("5. Generating from custom RNG:");
+    // 4. Generating from the custom StdRng instance
+    println!("4. Generating from custom RNG:");
     println!("  Random u32: {}", custom_std_rng.gen::<u32>());
     println!("  Random i32: {}", custom_std_rng.gen::<i32>());
-    println!("  Probability [0,1): {:.6}", custom_std_rng.gen_f64());
+    println!("  Probability [0,1): {:.6}", custom_std_rng.gen::<f64>());
     println!("  Range 1-100: {}", custom_std_rng.gen_range(1..=100));
     println!(
         "  Range 0.0-10.0: {:.3}",
@@ -70,17 +60,17 @@ fn main() {
     println!("  Boolean (25%): {}", custom_std_rng.gen_bool(0.25));
     println!();
 
-    // 6. Using a different seed for another StdRng instance
+    // 5. Using a different seed for another StdRng instance
     let seed = 456;
-    println!("6. Another StdRng instance with different seed {seed}:");
+    println!("5. Another StdRng instance with different seed {seed}:");
     let another_std_rng: RandomGenerator<StdRng> = RandomGenerator::new(seed);
     for i in 0..3 {
         println!("  {i}: {:.6}", another_std_rng.gen::<f64>());
     }
     println!();
 
-    // 7. Creating a custom RNG instance with a different generator
-    println!("7. Custom RNG instance with SmallRng with seed {seed}:");
+    // 6. Creating a custom RNG instance with a different generator
+    println!("6. Custom RNG instance with SmallRng with seed {seed}:");
     let secure_rng: RandomGenerator<SmallRng> = RandomGenerator::new(seed);
     for i in 0..3 {
         println!("  {i}: {:.6}", secure_rng.gen::<f64>());
--- a/phylo/examples/rng_examples.rs
+++ b/phylo/examples/rng_examples.rs
@@ -1,0 +1,89 @@
+use rand::rngs::{SmallRng, StdRng};
+
+use phylo::random::{
+    gen, gen_probability, gen_range, global_rng, init_rng, GlobalRng, RandomGenerator,
+};
+
+fn main() {
+    // 1. Using the global RNG (StdRng-based)
+    let seed = 42;
+    println!("1. Global RNG (StdRng) with seed {seed} - reproducible:");
+    init_rng(seed);
+    for i in 0..3 {
+        println!("  {i}: {:.6}", gen::<f64>());
+    }
+    // Reset and show reproducibility
+    init_rng(seed);
+    println!("After resetting to seed {seed}:");
+    for i in 0..3 {
+        println!("  {i}: {:.6}", gen::<f64>());
+    }
+    println!();
+
+    // 2. Generating from global RNG:
+    println!("2. Generating from global RNG:");
+    println!("  Random f64: {:.6}", global_rng().gen::<f64>());
+    println!("  Random u32: {}", global_rng().gen::<u32>());
+    println!("  Random i32: {}", global_rng().gen::<i32>());
+    println!("  Probability [0,1): {:.6}", global_rng().gen_f64());
+    println!("  Range 1-100: {}", global_rng().gen_range(1..=100));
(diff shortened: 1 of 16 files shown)


phylo/examples/rng_examples.rs
```diff
@@ -0,0 +1,79 @@
+use rand::rngs::{SmallRng, StdRng};
```
kalxed, line 1: I still have to see the extent to which these are used but just wanted to highlight they are non-portable meaning if any test result uses SmallRng even though it's a PRNG it might have different results on different platforms or in future versions

phylo/examples/rng_examples.rs
kalxed, line 1: I assume this file is simply for testing RNG examples though not automated? I think it makes more sense so people can see examples of random usage.

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -123,7 +124,12 @@ where
     /// assert_eq!(result.cost.tree().len(), 9); // The initial tree has 9 nodes, 5 leaves and 4 internal nodes.
     /// # Ok(()) }
     /// ```
-    pub fn run(mut self) -> Result<PhyloOptimisationResult<C>> {
+    pub fn run(self) -> Result<PhyloOptimisationResult<C>> {
+        let mut rng = DefaultGenerator::default();
+        self.run_w_rng(&mut rng)
+    }
+
+    pub fn run_w_rng(mut self, rng: &mut impl RandomSource) -> Result<PhyloOptimisationResult<C>> {
```
kalxed: This is cool, I assume just for testing purposes? I think overall good implementation with run having the default generator.

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -137,14 +143,7 @@ where
...
         let possible_prunes: Vec<_> = self.move_opti.move_locations(&self.c).copied().collect();
-        let current_prunes: Vec<_> = possible_prunes.iter().collect();
-        cfg_if::cfg_if! {
-        if #[cfg(not(feature = "deterministic"))] {
-            let mut current_prunes = current_prunes;
-            // TODO: decide on an explicit and consistent RNG to use throughout the project
-            let rng = &mut rand::thread_rng();
-        }
-        }
+        let mut current_prunes: Vec<_> = possible_prunes.iter().collect();
```
kalxed, line 148: So sauber! Great rework!

phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -12,6 +12,7 @@ use crate::optimisers::{
 use crate::parsimony::{scoring::ModelScoringBuilder, BasicParsimonyCost, DolloParsimonyCost};
 use crate::phylo_info::{PhyloInfo, PhyloInfoBuilder as PIB};
 use crate::pip_model::{PIPCost, PIPCostBuilder as PIPCB, PIPModel};
+use crate::random::FakeGenerator;
```
kalxed, line 16: General principle is that fake generator is used in every test file?

phylo/src/random/fake_random.rs
kalxed, line 1: Very nicely done, tests look good too!

phylo/src/random/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    fn rng_range() {
+        let rng = DefaultGenerator::new(123);
+        for _ in 0..10 {
+            let random_value: u32 = rng.gen_range(1..100);
+            assert!((1..100).contains(&random_value));
+        }
+    }
+
+    #[test]
+    fn different_seeds_produce_different_values() {
```
kalxed, line 242: Just want to highlight this is not necessarily true, though statistically probably will never happen. I think it's a fair test in the way that it's done. Unless implementation of StdRng changes (which is could) it shouldn't fail, and in that case you change the seeding values. Sorry if this is a completely unnecessary comment, as I don't think there is a better way to mitigate, or if it's even necessary.

phylo/src/random/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+        let _random_f64: f64 = rng.gen::<f64>();
+        let _random_probability = rng.gen_probability();
+        let _random_bool = rng.gen_bool(0.5);
+        assert!((0.0..1.0).contains(&rng.gen_probability()));
+    }
+
+    #[test]
+    fn rng_range() {
+        let rng = DefaultGenerator::new(123);
+        for _ in 0..10 {
```
kalxed, line 235: I was wondering for many of the tests whether the amount of times you do it matters? Can't hurt is my thought

phylo/src/random/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+
+/// Type alias for the default RNG implementation.
+/// Currently uses StdRng for good performance and reproducibility.
+pub type DefaultGenerator = RandomGenerator<StdRng>;
+
+impl Default for DefaultGenerator {
+    fn default() -> Self {
+        let seed = Timestamp::now().as_u64(); // Use current timestamp as seed
```
kalxed, line 92: I was curious and this will only be a problem after something like 600k years, how nice.

phylo/src/tree/mod.rs
```diff
@@ -411,28 +412,20 @@ fn argmin_wo_diagonal(q: Mat) -> (usize, usize) {
...
-    } else {
-        arg_min[rng_len(arg_min.len())]
-    }
-    }
-}
-
-#[cfg(not(feature = "deterministic"))]
-fn rng_len(l: usize) -> usize {
-    rand::random::<usize>() % l
+    arg_min[rng.gen::<usize>() % arg_min.len()]
```
kalxed, line 415: I checked this one for myself, and I got issues when generating a usize just off the rip from StdRng generator, but since there is the downcasting implemented I think it might be fine? Regardless I assume this basically generates a u64 to randomize the index?

kalxed (approved):
Looks great to me, will need to rework my code with this but overall a great implementation, there is a bit lost in the very close details but I meticulously checked the wiki with most of my confusion so I think everything makes sense

phylo/examples/rng_examples.rs
junniest: These are just there as an example, `SmallRng` is only used to show that it's possible to create a `RandomSource<SmallRng>`. Both `SmallRng` and `StdRng` are potentially non-portable from system to system, but the results for running anything with them should be still deterministic if the setup is fixed (i.e. the code is run on the same machine, same system version, same seed). Although I was planning to replace `StdRng` as the default generator and got stuck due to a dependency clash, thanks for reminding me to look into that again!

phylo/examples/rng_examples.rs
junniest: This is actually not for testing, it's an example binary that you can run with `cargo run --example rng_examples`. I didn't know cargo could do that until very recently, but it's a nice way to provide examples other than the doctests.

phylo/src/optimisers/topo_optimiser.rs
junniest: Good catch. This was a copy-paste from the NJ functionality, and this should be handled differently (imo the optimiser should be configured with a random source). I didn't do that in NJ because it would just make any merging more messy, but eventually we should also configure the NJBuilder with a RandomSource.

phylo/src/optimisers/topo_optimiser_tests.rs
junniest: Wherever we need a random generator but need the results to be deterministic, yes!

phylo/src/random/mod.rs
junniest: The only potential issue of using the timestamp as the seed is if two independent processes are started within the same microsecond, which would produce the same seed and so the same string of pseudo random numbers. But I find the risk of that acceptable :D

phylo/src/random/mod.rs
junniest: Good question. I have no idea, sadly (:  While googling this the only place I got to is a reddit rabbit hole, so I'm going to keep it as is for now.

phylo/src/random/mod.rs
junniest: I agree with you on this. It is definitely not guaranteed and it is possible that the generated values are the same. Though since the sampled values are f64, I think it is extremely unlikely.
I also can't see a better way to test this, but I definitely think that this test warrants a comment addressing your point.

phylo/src/tree/mod.rs
junniest: Hm, that's interesting, I can straight up generate usize from StdRng. The downcasting is only for the fake_generator so that we don't have to populate an infinite number of sources.

phylo/src/random/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+use rand::distributions::{Distribution, Standard};
+use rand::prelude::SliceRandom;
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+
+pub mod fake_random;
+pub use fake_random::*;
+
+/// Trait for random number generation
+pub trait RandomSource {
```
merlinio2000, line 14: I don't quite understand why you have to define a trait containing seed, gen, gen_bool and all that stuff.
Is your goal to be able to swap out  `rand` for another library at will?

Otherwise, I would suggest creating an extension trait only containing your specifically needed stuff like `gen_probablility`

phylo/src/random/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+///
+/// // Create a custom RNG instance
+/// let custom_rng: RandomGenerator<StdRng> = RandomGenerator::new(123);
+/// let custom_value: f64 = custom_rng.gen();
+/// ```
+pub struct RandomGenerator<R>
+where
+    R: Rng + SeedableRng + Send,
+{
+    r: Mutex<SeededRng<R>>,
```
merlinio2000, line 65: No me gusta, since the order when multithreading is out the window anyway (no reproducibility) I think it would make much more sense to work with some sort of threadlocal storage

phylo/src/random/mod.rs
junniest: Yes, the idea was to be able to use whatever rng people want, and also to swap out the default here to something platform-independent.

Unfortunately I didn't get it to work just yet because of some weird dependency clash, both `rand_pcg` and `rand_chacha` crates cause the clash. I think that the clashing dependency comes from `bio`, but I'll have to dig more to see what it is.

The other issue is that the way the standard rngs are seeded is different to what people normally use in comp bio. We tend to use a single number because it's easier to reproduce then and no one cares about the rng being crypto-secure, it just needs to be good enough. And for these kinds of applications I feel like we should be able to get the seed back e.g. for debug, which is not something that the `SeededRng` trait provides (for a good reason, but again, here we just need it to be good enough, not secure). I don't see how it would be possible without defining an additional structure like the `SeededRng` below, but please correct me if I'm wrong! (And that's a bad name for it too, since it's almost the same as `SeedableRng`, same for `RandomSource`...)

The other thing is that I defined a trait for this is so that I can make a fake generator that we can supply to tests. Implementing the `Rng` and `SeedableRng` for a fake was somehow very clunky so I ended up with this instead.

I am very certain that this is not the greatest design, just the best thing I could come up with in a limited time. Very happy to hear any suggestion for improving this.

phylo/src/random/mod.rs
junniest: Yes, totally agree. I was already thinking about that and at the time made the executive decision to not overthink it because a student was blocked by needing the `FakeRandom`.

The other thought here though is that we are not obliged to provide a perfectly reproducible version of a random generator that will handle threads in a reproducible way too. I am also not entirely sure it is possible to enforce either. 

In any case, at the moment the parallel parts of the spr move evaluation don't use any rng, rng is used to shuffle the pruning locations and then the threads work deterministically. Of course, if they ever need rng, this will be a different story, but I think I have a way in mind how to keep the reproducibility in the JATI binary without sacrificing parallel computing (supplying local randomisers seeded by values from the overall randomiser). 

Perhaps I should in general rename and comment on these structs to state clearly that this is not a general purpose random provider, but a very specifically limited version meant to promote computational reproducibility in phylogenetics. I still don't think that we can enforce completely deterministic behaviour on the crate users in general, so it might just be their individual responsibility to make sure their code is reproducible in the end.

Happy to hear your thoughts on this. This turned out surprisingly difficult to design in a meaningful way, and most likely should be refactored asap anyway (:

phylo/src/random/mod.rs
merlinio2000: Ah yes that totally makes sense, especially the seeding. Crytpographic random providers are definitely overkill for you. Just using unix/libc `rand` might work, but at the top of my head I don't know how that behaves in a multi threaded context.

The clashes between rand_ libraries and bio seem curious. I can't try it out right now but have you tried to import the rand libraries under an alias? Then there should definitely not be a conflict. In general a single crate can be instantiated multiple times with different version in the same build no problem.

About reproducibility: should this eventually be able to be 'deterministic'? In that case I thinks it unavoidable that you generate one giant sequence of random numbers and then distribute slices of values to threads in a deterministic order.

phylo/src/random/mod.rs
merlinio2000: I keep forgetting that people will also use this as a library not just a binary, thanks. Renaming this to be clear that its not meant for general purpose use (and maybe also make it private) seems like a good idea.

From a library design point, passing the generator explicitly around and not making it some global state is probably the most important part. Another thing about library usage, your RNG trait currently only takes immutable references to self, this is only made possible because Mutex allows interior mutability. I suggest making the trait take self by `&mut` (which `rand_` also does) as the generator carries an internal state that gets modified for each new number. If you need the shared references (e.g. passing to multiple threads) I would make that the users worry to properly solve as hes probably in dangerous territory anyways

phylo/src/random/mod.rs
junniest: Yes, the clash is very strange to be honest. I suspect that `bio` uses rand somewhere and defines something that depends on it, but I am legitimately only using the fasta reader from that crate which shouldn't be causing trouble... And I agree that there shouldn't be a dependency clash :D but for some reason cargo thinks otherwise and I've given up on messing with it for now, I am working on running the benchmarks at the moment. Obviously, our cluster setup is entirely different to what you guys have, so that's just great fun.

For reproducibility: I think that I ended up confusing the determinism in terms of computational reproducibility and the determinism needed for tests. The FakeGenerator is good enough for tests so far, and that's the only true determinism that we need. The version of determinism for computational reproducibility (and that only applies to the jati binary and the same version of that binary) just means that if I run the binary twice with the same seed I should get identical logs (with the exception of timestamps in the log). The good thing is that it actually seems to be working as intended with the current implementation, but I've only tested it manually so far (i.e. ran the thing twice with the seed of 1 and got the exact same numbers). I will add tests that verify this to the binary, but the library should not care about that, I think.

And yes, if we ever need the threads to also use something randomised, what you say seems like a good idea.

phylo/src/random/mod.rs
junniest: Yes, I will refactor this as soon as I have the time. Good call on the mutability, since the `Mutex` in any case will not make this reproducible if a bunch of threads also ask for randomisation. And I agree, I get mixed up in the library/binary specific requirements, but the binary requirements that I personally have should not apply to the library, it's the user's concern.

Unfortunately (or fortunately?), I specifically made a separate repo for the jati binary to also make sure that the library works as intended from an external crate. But this also means that the random generators cannot be private, since then I won't be able to use them either :D But I can use them as intended (and that way maybe also provide a usage example) and make sure that the same seeds generate the same results. As I mentioned in the other comment, on first glance it seems like it, and I think that's good enough for me.

PR #96, opened 2025-09-04
Enhancement/clean up optimisers

junniest:
Refactored the branch and model optimisation to remove unnecessary `RefCell` wrappers around cost objects.
Added max_iters to branch length optimisation to speed up long full tree optimisations.

### DIFF
--- a/phylo/src/optimisers/spr_optimiser.rs
+++ b/phylo/src/optimisers/spr_optimiser.rs
@@ -81,7 +81,7 @@ impl SprOptimiser {
         if tree.children(&tree.root).contains(prune_location) {
             // due to topology change the current node may have become the direct child of root
             // so the move is no longer possible
-            return Ok(MoveCostInfo::new(base_cost, tree.clone(), vec![]));
+            return Ok(MoveCostInfo::new(base_cost, tree.clone()));
         }
 
         let regraft_locations = self
@@ -215,7 +215,7 @@ fn calc_spr_cost_with_blen_opt<C: TreeSearchCost + Clone + Display>(
 ) -> Result<MoveCostInfo> {
     let mut new_tree = rooted_spr(cost_fn.tree(), &prune_location, &regraft)?;
 
-    cost_fn.update_tree(new_tree.clone(), &[prune_location, regraft]);
+    cost_fn.update_tree(new_tree.clone());
 
     let mut move_cost = cost_fn.cost();
     if cost_fn.blen_optimisation() && move_cost <= base_cost {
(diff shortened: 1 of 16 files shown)


phylo/src/optimisers/spr_optimiser.rs
kalxed, line 1: Is this basically so that dirty nodes are not redundantly kept track of within the struct and passed around in functions?

kalxed (approved):
I think this has probably been already thoroughly tested and checked enough but based on what I saw, you moved from vector of booleans to fixed bit set, which is nice change, and reworked how functions work with dirty nodes so that there isn't redundant data being passed around and for the most part it is stored in this bit set for all tree optimisers. Nice work!

phylo/src/tree/mod.rs
```diff
@@ -87,43 +88,29 @@ impl Tree {
...
+        })
+    }
+
+    pub fn clean(&mut self) {
+        self.dirty.set_range(0..self.dirty.len(), false);
     }
 
-    pub fn clean(&mut self, clean: bool) {
-        self.dirty.fill(clean);
+    pub fn dirty(&mut self) {
```
MattesMrzik, line 115: Could you add a single line doc comment to `clean()` and `dirty()`?

phylo/src/tree/tests.rs
```diff
@@ -47,7 +48,12 @@ fn setup_test_tree() -> Tree {
 #[test]
 fn single_leaf_tree_complete() {
     let sequences = Sequences::new(vec![record!("A0", b"AAAAAA")]);
-    let tree = Tree::new(&sequences).unwrap();
+    let mut tree = Tree::new(&sequences).unwrap();
+
+    tree.complete = true;
+    tree.compute_postorder();
+    tree.compute_preorder();
```
MattesMrzik, line 55: Do we perhaps instead want to have `tree.complete` be set by `Tree::new()`? And perhaps also have the `compute_postorder()` and `compute_preorder()` be called also by `Tree::new()`?

MattesMrzik (changes requested):
Looks good, could you pls add the doc that I mentioned below?

phylo/src/optimisers/spr_optimiser.rs
junniest: Yes, exactly. The other issue I noticed is that after an SPR move the dirty nodes set in the tree were actually different to the nodes being passed to the `update_tree()` function. I did not notice that before, but the nodes that were passed weren't correct, it was the regraft and the prune branch, whereas after an SPR move the prune sibling and the new prune parent were the ones that needed to be cleaned up. They could be recalculated, but it seemed unnecessary to do twice, since the dirty nodes are set correctly during the move. With the way it's implemented now there's only one source of information for this and fewer options for errors.

phylo/src/tree/tests.rs
junniest: Unfortunately, `new()` doesn't actually make a complete tree, unless the tree only has a single node. It creates a "template" with the right amounts of memory allocated and that template should be filled out by the tree builder. The post- and pre-order traversals can't be generated until all the nodes are properly added.

I don't like the boolean var `complete`, and would love to get rid of it, and make sure that the orders are properly created (there's no point in time when the tree is accessible but in this weird incomplete stage). However, at this point I am waiting for Kai to merge his PR so that I don't modify too much and make it difficult for him to merge. Once that is merged I will definitely clean this up more!

phylo/src/tree/mod.rs
junniest: Done and added tests for sanity.

MattesMrzik (approved):

Issue #47, opened 2025-04-14
Rerooting Tree


MattesMrzik:
Implement rerooting a tree. This could go in a module together with #32 and #31.

This rerooting should place the root right next to another node, i.e. with distance (blen) zero.

MattesMrzik:
Think about: 
> Should I refactor the tree to have three types of nodes instead of two? I.e. Internal, Leaf, Root? Since you only care about whether a node is the root or not. However, I might need to think about rerooting first and how that would affect the structure.

which was mentioned in #99

MattesMrzik:
If we are able to reroot the tree, then we can streamline the branch length optimization by rerooting to the edge that is being optimized. Since then we can reuse more of the internal tmp values from the logl calculation.

PR #99, opened 2025-09-17
Implementation of TKF92

### LINKED ISSUE #58: Implement TKF92 model
- implement the TKF92 cost and `ModelSearchCost` (without any reestimation of ancestral sequences)

MattesMrzik:
Implementation of the TKF92 cost for given tree and Multiple Ancestral Sequence Alignment (MASA), ie an alignment that includes ancestral wildcard sequences. 

Closes #58. Also implements TKF91 (since this is a special case of TKF92 when the parameter r = 0)

The functionality of fixing/re-estimating ancestral sequences after a tree move (see #115) will be addressed in another PR, which will include the impl of `TreeSearchCost`


phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -263,6 +264,7 @@ pub(crate) fn set_missing_tree_node_ids(tree: &Tree) -> Result<Tree> {
                 new_id = format!("I{count}");
             }
             tree_with_all_ids.nodes[usize::from(node_idx)].id = new_id;
+            info!("Set missing id of node {node_idx} to I{count}");
```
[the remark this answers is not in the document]
MattesMrzik: thats good, but it's missing the new_id.clone() on the previous line

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+            "TKF92 with lambda = {}, mu = {}, r = {}",
+            self.lambda(),
+            self.mu(),
+            self.r(),
+        )
+    }
+}
+
+/// For a detailed explanation of the TKF92 model and how the probabilities are calculated see:
+/// <coming paper>.
```
junniest: This will definitely get forgotten. I think it makes sense to at least add the original TKF paper here for now and add a TODO.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    }
+}
+
+impl<AA: AncestralAlignment, T: TKF> Display for TKFIndelCost<AA, T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.model)
+    }
+}
+
+fn valid_tkf_indel_parameters(lambda: f64, mu: f64, r: f64) -> Vec<f64> {
```
junniest: This is only for tkf92, right? Because tkf91 doesn't have the `r` parameter.

Now that I look at it, aren't `lambda` and `mu` independent of `r`? I think this needs to be split up in two at least, since `r` only applies to tkf92 and is checked independently anyway.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    }
+    if r == 0.0 {
+        warn!(
+            "Tried to set r to invalid value 0. It must be in (0, 1). Setting r to {}. Hint: r = 0 yields special case: TKF91 model, consider using that instead.",
+             DEFAULT_R
+        );
+        valid_r = DEFAULT_R;
+    } else if r <= 0.0 || r >= 1.0 {
+        warn!(
+            "Tried to set r to invalid value {}. It must be in (0, 1). Setting r to 0.5",
```
junniest: nitpick: The warning should rather be:
```
        warn!(
            "Tried to set r to invalid value {r}. It must be in (0, 1). Setting r to {DEFAULT_R}."
        );
```
Just on the off chance `DEFAULT_R` changes.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+        self.model.params()[param]
+    }
+
+    fn set_param(&mut self, i: usize, v: f64) {
+        if self.model.set_param(i, v) {
+            self.model_info.borrow_mut().valid.fill(false);
+        }
+    }
+
+    fn param_range(&self, param: usize) -> crate::likelihood::ParamRange {
```
junniest: nitpick: unnecessary `crate::likelihood::`

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+
+#[allow(clippy::upper_case_acronyms)]
+pub trait TKF: Clone + Display {
+    // TODO: it might be better for model optimisation to have parameter lambda and scale s = mu/lambda,
+    // because of the constraint that mu > lambda.
+    fn lambda(&self) -> f64;
+    fn mu(&self) -> f64;
+    // TKF91 has 2 parameters: lambda and mu, TKF92 has 3 parameters: lambda, mu and r.
+    fn params(&self) -> &[f64];
+    fn set_param(&mut self, i: usize, v: f64) -> bool;
```
junniest: I think that throughout we should rather have not `i` and `v` but something like `idx` and `value`, so the signature would be rather:
`fn set_param(&mut self, idx: usize, value: f64) -> bool;`
I noticed in the rest of the code that it was difficult to read the implementations with just the singular letters for var names.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+        Ok(TKFIndelCost {
+            model,
+            phylo: self.phylo.clone(),
+            model_info: RefCell::new(info),
+        })
+    }
+}
+#[derive(Clone, Debug)]
+pub struct TKFCost<Q: QMatrix + Display, T: TKF, AA: AncestralAlignment> {
+    // TODO: if we have just the sum of the two costs like this, we need to keep track of the
```
junniest, line 16: Good point, we should likely run flamegraphs to see if this is a big deal. It could be, but parallelisation requires the cloning to make sure every cost has valid info attached, which is why each cost instance has it's own clone.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+}
+
+fn n0(m: f64, b: f64) -> f64 {
+    m * b
+}
+
+fn log_n1(l: f64, m: f64, b: f64, t: f64) -> f64 {
+    ((1.0 - (-m * t).exp() - m * b) * (1.0 - l * b)).ln()
+}
+fn get_block_lens(blocks: &[usize]) -> Vec<usize> {
```
junniest: nitpick: newline before

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    let mut block_lens = vec![0; blocks.len()];
+    for (i, block) in blocks.iter().enumerate() {
```
junniest: I am not sure what's going on here. Are you computing block lengths or block boundaries? I suggest renaming this to explain what the lens is since this is not a common (?) term and potentially adding a comment.

phylo/src/optimisers/model_optimiser_tests.rs
```diff
@@ -574,3 +576,53 @@ fn stop_condition_epsilon() {
...
+    let subst_model = SubstModel::<HKY>::new(&[], &[2.0]);
+    let tkf91 = TKF91CostBuilder::new(0.8, 1.0, subst_model.clone(), info.clone())
+        .build()
+        .unwrap();
+    let tkf92 = TKF92CostBuilder::new(0.8, 1.0, 0.2, subst_model, info)
+        .build()
+        .unwrap();
+
+    tkf_model_opti_template(tkf91);
+    tkf_model_opti_template(tkf92);
```
junniest: I suggest splitting these into two tests after all. Unfortunately templates make it a bit difficult to see which of the template calls failed, so having `tkf91_model_opimisation` and `tkf92_model_opimisation` tests would be more helpful to know where the error happened.

phylo/src/optimisers/spr_optimiser.rs
```diff
@@ -218,6 +218,9 @@ fn calc_spr_cost_with_blen_opt<C: TreeSearchCost + Clone + Display>(
     cost_fn.update_tree(new_tree.clone());
 
     let mut move_cost = cost_fn.cost();
+    // If the move_cost is already better, then we dont need to call the blen optimiser
```
junniest: Since you've already added this comment: some models don't have branch lengths (parsimony), so to make this comment exhaustive I suggest rephrasing to:
```
// Branch length optimisation is skipped if the model does not support branch lengths 
// or if the move_cost is already better since we apply the move as is.
```
The current comment is also not strictly correct because after applying all moves we reoptimise all branch lengths, but in this case we only try to optimise the regraft branch.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    static ref DUMMY_FREQS: DVector<f64> = DVector::<f64>::zeros(0);
+}
+
+static DEFAULT_LAMBDA: f64 = 1.0;
+static DEFAULT_MU: f64 = 1.1;
+static DEFAULT_LAMBDA_MU_RATIO: f64 = 0.9;
+static DEFAULT_R: f64 = 0.5;
+
+#[allow(clippy::upper_case_acronyms)]
+pub trait TKF: Clone + Display {
```
junniest: I would maybe rename this trait to `TKFModel` just to match the other trait definitions.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    }
+
+    fn params(&self) -> &[f64] {
+        &self.params
+    }
+
+    /// Sets the parameter if it is valid then returns true, otherwise the parameter is not changed
+    /// and false is returned.
+    /// This assumes that the other parameter is valid
+    fn set_param(&mut self, i: usize, v: f64) -> bool {
```
junniest: As discussed, changing `i` -> `idx` and `v` -> `value` would be helpful, and the parameter enum would help readability.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+
+    fn get_blocks<AA: AncestralAlignment>(msa: &AA) -> Vec<usize> {
+        let mut blocks: HashSet<usize> = HashSet::new();
+        for map in msa
+            .ancestral_maps()
+            .values()
+            .chain(msa.leaf_maps().values())
+        {
+            let mut last = map[0].is_some();
+            for (i, c) in map.iter().skip(1).enumerate() {
```
junniest: This is quite hard to read, could you consider renaming the variables to provide context?

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    /// h1[node] = h1(node.blen), may hold previously computed values for h1 that can be reused.
+    h1: Vec<Option<f64>>,
+
+    /// insertion[node] = l * beta[node] * (1.0 - r) / r, may hold previously computed values for insertion
+    /// that can be reused.
+    insertion: Vec<Option<f64>>,
+
+    /// factor_n[node] = n1/ (n0 * lambda * beta(node.blen)), may hold previously computed
+    /// values for factor_n that can be reused.
+    factor_n: Vec<Option<f64>>,
```
junniest: Are all 4 of these valid/invalid together? If so, I would suggest using a FixedBitSet to mark valid/invalid values instead of having Options here. I might be wrong, but it seems to me that the constant conversion between option and value and the checks for values (4 times, for each of these) will add up time-wise. I am still using Vec<bool> in PIP and substitution models to keep track of valid/invalid values, but I think that FixedBitSet's should be better for this purpose.

junniest: If you end up using flags, these can also potentially be stored as DMatrix'es (or DVectors) just for consistency.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+
+    fn set_param(&mut self, i: usize, v: f64) {
+        if self.model.set_param(i, v) {
+            self.model_info.borrow_mut().valid.fill(false);
+        }
+    }
+
+    fn param_range(&self, param: usize) -> crate::likelihood::ParamRange {
+        match param {
+            0 => (f64::EPSILON, self.model.mu()),
```
junniest: This would also be refactored with a parameter def enum.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+            1 => (self.model.lambda(), f64::MAX),
+            2 => PARAM_RANGE_UNIT_INTERVAL_EXCLUSIVE,
+            _ => panic!("Invalid parameter index for TKF model: {}", param),
+        }
+    }
+
+    fn set_freqs(&mut self, _: FreqVector) {}
+
+    fn empirical_freqs(&self) -> FreqVector {
+        // TODO: At the time of writing this, this method is only used to set the frequencies of
```
junniest: I think it's better to leave it as is, since the empirical frequencies are defined by the data, and it does not truly matter if the model has them or not.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+        // the model, but the TKF92IndelCost does not have frequencies. So we could just return
+        // a dummy vector here.
+        self.phylo.freqs()
+    }
+
+    fn freqs(&self) -> &FreqVector {
+        // TODO: Alternatively, we don't implement costs for just the indel part, or we make the
```
junniest: Good q. But we could, theoretically, run tree search with just the indel costs? Not sure why we would want to, but we could. I think dummy freqs are fine for this and I wouldn't bother with changing the trait definition.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+                .get_or_insert_with(|| self.model.insertion_prob_at_root());
+        }
+        1.0
+    }
+
+    fn get_indel_x_and_factor_n_for_non_root(
+        &self,
+        node_idx: &NodeIdx,
+        block_id: usize,
+    ) -> (f64, f64) {
```
junniest: I guess there's no way to decouple the computation of x and n, right? I am always worried about returning tuples/vectors of values because it makes it increasingly easy to forget which value is which, especially when they're both the same type. Not sure this warrants a newtype definition, but still something to consider -- at least maybe adding a comment to state which value is which?

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    (-m * t).exp() * (1.0 - l * b)
+}
+
+fn n0(m: f64, b: f64) -> f64 {
+    m * b
+}
+
+fn log_n1(l: f64, m: f64, b: f64, t: f64) -> f64 {
+    ((1.0 - (-m * t).exp() - m * b) * (1.0 - l * b)).ln()
+}
```
junniest: here again I think it would be helpful to rename single letters into their appropriate names (I guess l is always lambda and m is always mu?). Otherwise it gets very confusing.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+
+    // act
+    let blocks = TKF92IndelModel::get_blocks(&msa);
+    let block_lens = get_block_lens(&blocks);
+    assert_eq!(blocks, vec![1, 3, 4, 5]);
+    assert_eq!(block_lens, vec![1, 2, 1, 1]);
+}
+
+#[cfg(test)]
+fn tkf_set_lambda_and_mu<T: TKF>(model: &mut T) {
```
junniest: I think this and the following test funcs will be easier to read with the enum for parameter definitions. I would also potentially split lambda and mu setting into two separate functions.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+        TKF92CostBuilder::new(1.0, 2.0, 0.3, subst_model, setup_test_phylo(dna_alphabet()))
+            .build()
+            .unwrap();
+    assert_eq!(tkf_cost.freqs().as_slice(), &[0.1, 0.2, 0.3, 0.4]);
+    tkf_cost.set_freqs(frequencies!(&[0.4, 0.3, 0.2, 0.1]));
+    assert_eq!(tkf_cost.freqs().as_slice(), &[0.4, 0.3, 0.2, 0.1]);
+}
+
+#[test]
+fn tkf91_indel_logl_() {
```
junniest: nitpick: stray `_` at the end of name

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+    // assert
+    assert_relative_eq!(logl, manual_calculation);
+    assert_relative_eq!(logl, half_manual);
+}
+
+#[test]
+fn tkf_cost_builder_fails() {
+    // arrange
+    let phylo = setup_test_phylo(protein_alphabet());
+    let subst_model = SubstModel::<GTR>::new(&[0.1, 0.3, 0.4, 0.2], &[1.2, 0.5, 5.0, 1.0, 1.0]);
```
junniest: Just as a comment: you can also skip providing all the parameters if it's irrelevant, so you can do:
`SubstModel::<GTR>::new(&[], &[]);` here and in all other places.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+        .build()
+        .unwrap_err()
+        .to_string();
+    let tkf92_err_msg = TKF92CostBuilder::new(0.1, 0.2, 0.3, subst_model, phylo)
+        .build()
+        .unwrap_err()
+        .to_string();
+
+    // assert
+    assert_eq!(
```
junniest, line 628: nitpick: Here again potentially could be helpful to split this in two tests so you know for sure which of the models fails.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+    // act
+    let logl = tkf_cost.cost();
+    let subst_logl = subst_cost.cost();
+    let tkf_logl = tkf91_indel_logl_without_aggregation(
+        &tkf_cost.indel_cost.model,
+        &tkf_cost.indel_cost.phylo,
+    );
+
+    // assert
+    assert_relative_eq!(logl, subst_logl + tkf_logl, epsilon = 1e-12);
```
junniest: question: Is there a reason why the costs would not match exactly?

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+    );
+
+    // assert
+    assert_relative_eq!(tkf_indel_cost_1, tkf_indel_cost_without_agg_1);
+    assert_relative_eq!(tkf_indel_cost_2, tkf_indel_cost_without_agg_2);
+    assert_relative_eq!(tkf_log_1 - tkf_indel_cost_1, tkf_log_2 - tkf_indel_cost_2);
+}
+
+#[cfg(test)]
+fn setup_test_phylo(alphabet: Alphabet) -> PhyloInfo<MASA> {
```
junniest: nitpick: Would be helpful to move the setup funcs to the beginning of the file (or before they are first used)

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+    let tkf_cost = TKF92CostBuilder::new(0.1, new_tkf_mu, 0.3, subst_model, phylo)
+        .build()
+        .unwrap();
+    let new_logl = ModelSearchCost::cost(&tkf_cost);
+    assert_eq!(new_logl, ModelSearchCost::cost(&tkf_cost));
+    assert_eq!(logl2, new_logl);
+}
+
+#[test]
+fn tkf_modify_model_params_costs_match() {
```
junniest: nitpick: again, I think these should probably be 3 separate tests, and the name should start with tkf92 since this does not apply to tkf91.

junniest (commented):
This is quite hefty to review, so I think we'll need another round.
One more thought: I think it might be worth at this point to split the module into at the very least main (`mod.rs`), `tkf91_model` and `tkf92_model` or something along these lines (and `tkfindel_model`?). It becomes quite difficult to figure out which model is which in a file this long.

phylo/src/tkf_model/mod.rs
MattesMrzik: also just found this:)

phylo/src/tkf_model/mod.rs
MattesMrzik: what about: 
```rust
    // Determines the block borders from the alignment. A block border is defined as a
    // position where any sequence changes from gap to non-gap or vice versa. Returns a sorted
    // vector of the right exclusive block borders.
    fn get_blocks<AA: AncestralAlignment>(msa: &AA) -> Vec<usize> {
        let mut blocks: HashSet<usize> = HashSet::new();
        for map in msa
            .ancestral_maps()
            .values()
            .chain(msa.leaf_maps().values())
        {
            let mut previous_is_char = map[0].is_some();
            for (i, c) in map.iter().skip(1).enumerate() {
                let current_is_char: bool = c.is_some();
                // whenever there is a change from gap to not gap or vice versa, we have a block border
                if previous_is_char != current_is_char {
                    blocks.insert(i + 1);
                }
                previous_is_char = current_is_char;
            }
            blocks.insert(map.len());
        }
        let mut blocks: Vec<usize> = blocks.iter().copied().collect();
        blocks.sort();
        blocks
    }

```

phylo/src/tkf_model/mod.rs
MattesMrzik: There might be for example an index where the value for `h1` is valid but the `factor_n` not, this would mean having 4 DVectors that keep track of which values are calculated or not. I think for the majority of MSAs (especially if not many indel, or indels always happen at similar nodes) then we might not need the values for `factor_n` for example for every node. So i was thinking that unwrapping always but potentially having to calculate `factor_n` (which included exponential) for only a small subset of nodes.

But now that i think of that, it might only be worth in the case for `factor_n`, since for the others its either expected that we need the value for every node any way, or its just one simple multiplication. In the case of the `insertion` i dont know what is faster: either having to unwrap if we encounter an insertion, or having to calculate `beta[node] * (1.0 - r) / r` (a single unwrap is probably faster than 4 multiplications?)

phylo/src/tkf_model/mod.rs
MattesMrzik: I was just thinking that calling this would be a waste of time if the result is not needed anyway.

phylo/src/tkf_model/mod.rs
MattesMrzik: Yep, i would also not know why we wound want to run tree search with just indels, but it might be nice to test it.

phylo/src/tkf_model/mod.rs
MattesMrzik: No, i can decouple it:)
I would extract this snipped: 
```rust
let parent_idx = self.phylo.tree.node(node_idx).parent.unwrap();
        let node_id = usize::from(node_idx);
        let mut factor_n = 0.0;
        let mut x: f64 = 1.0;
        let site = self.model_info.borrow().blocks[block_id] - 1;
        let parent_is_gap = match parent_idx {
            Internal(_) => self.phylo.msa.ancestral_map(&parent_idx)[site].is_none(),
            _ => unreachable!("The parent of a node cannot be a leaf."),
        };
        let current_is_gap = match node_idx {
            Internal(_) => self.phylo.msa.ancestral_map(node_idx)[site].is_none(),
            Leaf(_) => self.phylo.msa.leaf_map(node_idx)[site].is_none(),
        };
```
to another method that returns a enum Action, which can be Homolog, Insertion, Deletion, Nothing. (i am using the enum already in the unpublished reestimate ancestors code)

phylo/src/tkf_model/mod.rs
MattesMrzik: would adding this
```
/// Given the right exclusive block borders, returns the lengths of the blocks.
/// For example, given [3, 5, 8], the block lengths are [3, 2, 3].
```
be sufficient to make it clear?

phylo/src/tkf_model/mod.rs
MattesMrzik: makes sense, thanks for the remark. i split it into two methods.

phylo/src/tkf_model/mod.rs
junniest: Aaaaah, my brain read "lens" as a singular noun :D So I was trying to understand what do lenses have to do with blocks and if it's some fancy term for a view.
Maybe to avoid this confusion it still makes sense to call the function `get_block_lengths`? It's not too many more letters, but at least I won't read it weird...
In any case, yes, the comment would help regardless.

phylo/src/tkf_model/mod.rs
junniest: Yes, I think this is better! 
The only thing I would suggest changing (it's strange that clippy doesn't suggest this actually) is to use the XOR operation rather that `!=`, because `!=` makes it seem like you are comparing values instead of bools. Usually clippy/check point this out as an issue.

nitpick: this line: `let current_is_char: bool = c.is_some();` doesn't need the type annotation, so it should rather be `let current_is_char = c.is_some();`.

phylo/src/tkf_model/mod.rs
junniest: True. But I think this should still return valid stationary frequencies from the data just for consistency. This method could maybe be refactored out since it's not strictly a model property, but `ModelOptimiser` has no other way to get empirical frequencies because the `PhyloInfo` (still hate this name) is hidden inside the model cost... Not sure what to do about this, probably leave it as is for now.

phylo/src/tkf_model/mod.rs
junniest: It's the same for parsimony to a degree, it's possible to do tree search under it for consistency, but I wouldn't recommend doing that for phylogenetics in general.
One potential use case though could be as a faster initial tree search? Since I'm guessing that this isn't exponential (?) and would therefore be compatible with SPR moves?..

phylo/src/tkf_model/mod.rs
junniest: That sounds great! And you know I like meaningfully named enums haha.

phylo/src/tkf_model/mod.rs
junniest: I see. Yes, 4 BitSets are probably too much, but one could still be worth it (you'd use those for flags, not DVectors, those cannot store bools I think). I agree, I think that one unwrap should definitely be faster than 3-4 multiplications.

Also, if you're using `(1.0 - self.r()) / self.r()` a lot, could it be worth precomputing that? I'm guessing it would change upon model optimisation, but wouldn't change at all during tree search. It's not a lot of computation for sure, but at least one division can be removed.

phylo/src/tkf_model/mod.rs
MattesMrzik: hihi:)
yes sure. will rename to `get_block_lengths`

phylo/src/tkf_model/mod.rs
MattesMrzik: thanks:)

phylo/src/tkf_model/mod.rs
MattesMrzik: Actually, now I am much happier about how the code is structured. Thanks for pointing that out.

phylo/src/tkf_model/mod.rs
MattesMrzik: Thant's a good remark. I will rethink that.

phylo/src/tkf_model/mod.rs
MattesMrzik: It's the indel process that makes it hard, so we must stick to this co-estimation approach.

phylo/src/tkf_model/tests.rs
MattesMrzik: I split them and did 
```rust
fn tkf_set_mu<T: TKFModel>(model: &mut T) {
    let initial_lambda = model.lambda();
    let initial_mu = model.mu();
    let param_idx = usize::from(TKF92Parameters::Mu);
    assert!(!model.set_param(param_idx, -0.1));
```
is it better like this?

phylo/src/tkf_model/tests.rs
MattesMrzik: Mmh, i removed it, and it still passes, which is good news. I thought at some point there was a tiny discrepancy, but I guess not anymore.

phylo/src/optimisers/topo_optimiser.rs
```diff
@@ -30,6 +31,8 @@ impl<S: ParsimonyScoring, A: Alignment> Compatible<SprOptimiser> for DolloParsim
 impl<S: ParsimonyScoring, A: Alignment> Compatible<NniOptimiser> for DolloParsimonyCost<S, A> {}
 impl<A: Alignment> Compatible<SprOptimiser> for BasicParsimonyCost<A> {}
 impl<A: Alignment> Compatible<NniOptimiser> for BasicParsimonyCost<A> {}
+impl<AA: AncestralAlignment, T: TKFM> Compatible<NniOptimiser> for TKFIndelCost<AA, T> {}
+impl<Q: QMatrix, T: TKFM, AA: AncestralAlignment> Compatible<NniOptimiser> for TKFCost<Q, T, AA> {}
```
junniest: I feel like there's something strange going on with the order of generic parameters here. We should try to keep the order at least somewhat consistent, it's not always possible, but would be helpful.
That would mean that for `TKFIndelCost` it should be flipped:
`impl<T: TKFM, AA: AncestralAlignment> Compatible<NniOptimiser> for TKFIndelCost<T, AA> {}`
Then, roughly, the order for all of these will be Q (if necessary), model (if necessary), alignment.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+
+// TODO: link our paper once it is published. For now see original TKF92 paper: https://doi.org/10.1007/bf00163848
+#[derive(Clone, Debug)]
+struct TKFIndelModelInfo {
+    /// aggregated_x[(node, block)] = the product of the xs of all the edges in the subtree below
+    /// <node> (including the x of <node> itself) for the block with id <block>.
+    aggregated_x: DMatrix<f64>,
+
+    /// node_x[(node, block)] = the x value for the edge above <node> for the block with id <block>.
+    node_x: DMatrix<f64>,
```
junniest: What does `x` stand for? I saw in your manuscript (def 4.16) that it is the edge event factor (?) but I am still not sure what that means. Especially since this is new terminology, I think you should also mention/explain what the factor means in here, not only in the paper. Additionally, the exact terminology in the paper may change, so having a coherent description here would still be helpful.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    /// n0[node] = n0(node.blen)
+    n0: Vec<f64>,
+
+    /// h1[node] = h1(node.blen)
+    h1: Vec<f64>,
+
+    /// insertion[node], may hold previously computed values for insertion
+    /// that can be reused.
+    insertion: Vec<f64>,
+
```
junniest: You could remove this empty line (and others) to group the values and their validity descriptions together. That would help the logical flow of connected elements.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    /// node_x[(node, block)] = the x value for the edge above <node> for the block with id <block>.
+    node_x: DMatrix<f64>,
+
+    /// node_factor_n[(node, block)] = the factor_n value for the edge above <node> for the block
+    /// with id <block>.
+    node_factor_n: DMatrix<f64>,
+
+    /// factor_ns[(node, block)] = n1/ (n0 * lambda * beta(v.blen)) if there is a node <v> in the
+    /// subtree rooted in <node> where the current event is an insertion and the previous one was a deletion.
+    factor_ns: DMatrix<f64>,
```
junniest: The naming scheme here is confusing. Both these matrices seem to hold `factor_n` in some shape or form, but I cannot tell if `factor_ns` is an aggregator or `node_factor_n` or something else entirely. Also, `n` does not feature in your manuscript, but did you mean def 4.17? "η" is actually "eta" in Greek, and I cannot seem to find a definition for an "n" factor.

junniest: Or did you mean `N1` here?.. Then it is also quite confusing because `n0` and `h1` appear later, and also without a note on the meaning.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    /// insertion[node], may hold previously computed values for insertion
+    /// that can be reused.
+    insertion: Vec<f64>,
+
+    /// insertion_value_valid[node] = true if insertion[node] holds a valid value.
+    insertion_value_valid: FixedBitSet,
+
+    /// factor_n[node] = n1/ (n0 * lambda * beta(node.blen)), may hold previously computed
+    /// values for factor_n that can be reused.
+    factor_n: Vec<f64>,
```
junniest: One thing for sure: using `factor_ns` and `factor_n` is a recipe for disaster, imo (: The names are too similar and it's not immediately clear what the difference between the two is.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    blocks: Vec<usize>,
+
+    /// The lengths of the blocks.
+    block_lengths: Vec<usize>,
+
+    /// last_event_deletion[node] = true if the last event was a deletion for a that <node>.
+    previous_action_deletion: Vec<bool>,
+
+    /// valid[node] = true if the intermediate values for that <node> are valid.
+    valid: Vec<bool>,
```
junniest: Can/should also be a `FixedBitSet`

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+                        self.set_non_root(node_idx);
+                    }
+                }
+                Leaf(_) => {
+                    self.set_non_root(node_idx);
+                }
+            };
+        }
+
+        let l: f64 = self.model.lambda();
```
junniest: nitpick: unnecessary `f64`

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.model)
+    }
+}
+
+impl<AA: AncestralAlignment, T: TKFModel> TKFIndelCost<AA, T> {
+    fn logl(&self) -> f64 {
+        for node_idx in self.phylo.tree.postorder() {
+            match node_idx {
+                Internal(_) => {
```
junniest: Should I refactor the tree to have three types of nodes instead of two? I.e. `Internal`, `Leaf`, `Root`? Since you only care about whether a node is the root or not. However, I might need to think about rerooting first and how that would affect the structure.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    fn update_previous_event(&self, node_idx: &NodeIdx, action: Action) {
+        let node_id = usize::from(node_idx);
+        match action {
+            Action::Deletion => {
+                self.model_info.borrow_mut().previous_action_deletion[node_id] = true;
+            }
+            Action::Insertion | Action::Homolog => {
+                self.model_info.borrow_mut().previous_action_deletion[node_id] = false;
+            }
+            Action::Nothing => {}
```
junniest: Wouldn't `Action::Nothing` also mean that `previous_action_deletion[node_id] = false`? Just for consistency's sake, and on the off chance that value gets read by accident. Since now it would just remain as whatever was set previously, which turns invalid, if I understand this right.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+        self.model_info.borrow_mut().previous_action_deletion[node_id] = false;
+        self.model_info
+            .borrow_mut()
+            .insertion_value_valid
+            .set(node_id, false);
+        self.model_info
+            .borrow_mut()
+            .factor_n_value_valid
+            .set(node_id, false);
+        self.model_info.borrow_mut().valid[node_id] = false;
```
junniest: You will probably be better off here borrowing the whole info once and then updating it without needing to write `self.model_info.borrow_mut()` in every line:
```
let mut model_info = self.model_info.borrow_mut();
model_info.beta[node_id] = beta;
model_info.n0[node_id] = n0(mu, beta);
model_info.h1[node_id] = h1(lambda, mu, beta, blen);
model_info.previous_action_deletion[node_id] = false;
...
```
You can also use `drop(model_info)`, e.g. if you need to reborrow unmutably somewhere in the same function, but like this `model_info` will go out of scope at the end of the method and the mutable borrow will be released.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    fn set_node_values(&self, node_idx: &NodeIdx, block_id: usize, mut x: f64, mut factor_n: f64) {
+        let node_id = usize::from(node_idx);
+        self.model_info.borrow_mut().node_x[(node_id, block_id)] = x;
+        self.model_info.borrow_mut().node_factor_n[(node_id, block_id)] = factor_n;
+        for child in &self.phylo.tree.node(node_idx).children {
+            let child_id = usize::from(child);
+            x *= self.model_info.borrow().aggregated_x[(child_id, block_id)];
```
junniest: Here I would suggest creating a new var to indicate that it's not just `x` or `factor_n`, but an aggregated value, i.e.
```
let aggr_x = x;
for child in &self.phylo.tree.node(node_idx).children {
            let child_id = usize::from(child);
            aggr_x *= self.model_info.borrow().aggregated_x[(child_id, block_id)];
            ...
}
self.model_info.borrow_mut().aggregated_x[(node_id, block_id)] = aggr_x;
```

This would also remove the mutability from the signature of the method which adds confusion in my eyes. Copying/creating single f64 values has very little overhead (and will probably be optimised away by the compiler), so it seems unnecessary. We should try to limit copying/cloning vectors/matrices unnecessarily, but in this case a separate aggregator var will make the code easier to read.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+            factor_n += self.model_info.borrow().factor_ns[(child_id, block_id)];
+        }
+        self.model_info.borrow_mut().factor_ns[(node_id, block_id)] = factor_n;
+        self.model_info.borrow_mut().aggregated_x[(node_id, block_id)] = x;
+    }
+
+    fn get_indel_x_for_root(&self, block_id: usize) -> f64 {
+        let root_idx = &self.phylo.tree.root;
+        let site = self.model_info.borrow().blocks[block_id] - 1;
+        let char_at_root = self.phylo.msa.ancestral_map(root_idx)[site].is_some();
```
junniest: Maybe this would be more meaningfully named as `char_present_at_root`? Again, just calling something a `char` to me reads as if you need the value, but it's a bool variable.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+                    .set(usize::from(root_idx), true);
+                return insertion_prob;
+            }
+        }
+        1.0
+    }
+
+    fn get_action(&self, node_idx: &NodeIdx, block_id: usize) -> Action {
+        if block_id == 0 {
+            self.model_info.borrow_mut().previous_action_deletion[usize::from(node_idx)] = false;
```
junniest: This does not sit right with me. The method name is `get_action` implying that it returns something (an action). But at the same time it is setting something internally, creating a side effect. This can be extremely difficult to debug, so I would recommend doing the setting elsewhere.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+            if self
+                .model_info
+                .borrow()
+                .insertion_value_valid
+                .contains(usize::from(root_idx))
+            {
+                return self.model_info.borrow().insertion[usize::from(root_idx)];
+            } else {
+                let insertion_prob = self.model.insertion_prob_at_root();
+                self.model_info.borrow_mut().insertion[usize::from(root_idx)] = insertion_prob;
```
junniest: I wrote this in the comment below, but I will say it here again. If a method is called `get_something`, I expect it to return a value, which it does. What I do not expect is that it will also do some setting of values, which creates a side effect that can be difficult to debug and trace. Since you are using the value when you call this method, I would rather recommend renaming this to `set_something`, not return anything and then accessing the value from the `model_info` directly.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+            let mu = self.model.mu();
+            let beta = self.model_info.borrow().beta[node_id];
+            let blen = self.phylo.tree.node(node_idx).blen;
+
+            let n0_option = self.model_info.borrow().n0[node_id];
+            if self
+                .model_info
+                .borrow()
+                .factor_n_value_valid
+                .contains(node_id)
```
junniest: Here, again, I would recommend doing something like:
```
let model_info = self.model_info.borrow();
```
to remove the 5-line call.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    fn empirical_freqs(&self) -> FreqVector {
+        // TODO: At the time of writing this, this method is only used to set the frequencies of
+        // the model, but the TKF92IndelCost does not have frequencies.
+        self.phylo.freqs()
+    }
+
+    fn freqs(&self) -> &FreqVector {
+        // TODO: Alternatively, we don't implement costs for just the indel part, or we make the
+        // trait def to return an Option<&FreqVector>.
+        &DUMMY_FREQS
```
junniest: Or, this method could just return empirical frequencies as well to stay consistent. I am not sure that changing the signature here would be helpful since for most other models it is not an option.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    fn cost(&self) -> f64 {
+        self.indel_cost.cost() + self.subst_cost.cost()
+    }
+
+    fn param_count(&self) -> usize {
+        self.indel_cost.model.params().len() + self.subst_cost.model.qmatrix.params().len()
+    }
+
+    fn param(&self, idx: usize) -> f64 {
+        let num_params_indel_model = self.indel_cost.param_count();
```
junniest, line 44: This isn't very elegant, but I guess it gets the job done... I wonder if we could do this more elegantly, but I guess since you are implementing TKFCost for any of the possible models, there isn't a way to name/enumerate parameters better.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+fn b(lambda: f64, mu: f64, time: f64) -> f64 {
+    let exp_term = ((lambda - mu) * time).exp();
+    (1.0 - exp_term) / (mu - lambda * exp_term)
+}
+
+fn h1(lambda: f64, mu: f64, beta: f64, time: f64) -> f64 {
+    (-mu * time).exp() * (1.0 - lambda * beta)
+}
+
+#[inline]
```
junniest: Would it make sense to inline the other small functions too?

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+        self.subst_cost.freqs()
+    }
+}
+
+fn log_i1(lambda: f64, beta: f64) -> f64 {
+    (1.0 - lambda * beta).ln()
+}
+
+// TKF beta function
+fn b(lambda: f64, mu: f64, time: f64) -> f64 {
```
junniest: I find it curious that this is the one function which has a comment, and its name does not match it :D I would suggest renaming this to `beta` (although I don't remember if the original TKF definition has it).

I think that adding comments to all these functions would be helpful, same as for the internal values in `model_info`. Even if everything will be described in your manuscript in exactly the same way, comments would simplify reading the code.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+    node: &'a NodeIdx,
+) -> &'a Mapping {
+    match node {
+        Internal(_) => msa.ancestral_map(node),
+        Leaf(_) => msa.leaf_map(node),
+    }
+}
+
+#[cfg(test)]
+fn tkf91_indel_logl_without_aggregation<AA: AncestralAlignment>(
```
junniest, line 40: I would suggest adding a comment to this and the next function to indicate that this is the same computation written out for comparison purposes.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+}
+
+#[test]
+fn tkf_log_i1() {
+    // arrange
+    let l = 2.0;
+    let m = 3.0;
+    let time = 1.0;
+    let b = b(l, m, time);
+    // act & assert
```
junniest: As much as I understand why you are adding these comments (`arrange`, `act and assert`), I think that at least here they are unnecessary. It's already clear which part is the setup and which is the assertion and these comments don't clarify the test.
I would also suggest flipping the order of the assertion and the comment describing its value so that it's clear this is the value you expect and not an afterthought.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+fn tkf91_get_blocks() {
+    // arrange
+    let tree = tree!("((A0:1.0,B1:1.0)I1:1.0);");
+    let seqs = Sequences::new(vec![
+        record!("A0", b"AAAB-D"),
+        record!("B1", b"--ARAW"),
+        record!("I1", b"AAAA-A"),
+    ]);
+    let msa = MASA::from_aligned_with_ancestral(seqs, &tree).unwrap();
+    // act
```
junniest: I think that a newline instead of the comment would work for the same purpose without adding more text to read.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+    assert!(!model.set_param(param_idx, 2.1));
+    assert_eq!(model.lambda(), initial_mu - 0.001);
+}
+
+#[cfg(test)]
+fn tkf_set_mu<T: TKFModel>(model: &mut T) {
+    let initial_lambda = model.lambda();
+    let initial_mu = model.mu();
+    let param_idx = usize::from(TKF92Parameters::Mu);
+    assert!(!model.set_param(param_idx, -0.1));
```
junniest: I wonder if now that you have implemented parameter ranges it is still necessary for this method to return true/false. This won't get checked in the model optimiser anyway since it relies on the trait definition and now seems unnecessary.
My other, repeated, issue with this is that I expect a setter to set something and not return a value. I would rather it not have a return and rely on the valid parameter range definition to set the parameters.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+    // arrange
+    let mut tkf91_indel_model =
+        TKF91IndelCostBuilder::new(1.0, 2.0, setup_test_phylo(dna_alphabet()))
+            .build()
+            .unwrap()
+            .model;
+
+    // act & assert
+    tkf_set_lambda(&mut tkf91_indel_model.clone());
+    assert!(!tkf91_indel_model.set_param(2, 0.5)); // no r parameter
```
junniest: The other issue I feel is that for the other tkf models `set_param(2, 0.5)` is perfectly valid, since if it is used on TKF91, it will set a substitution model parameter to 0.5 (if there are any). So this feels like fairly random specific behaviour which we do not account for in most places.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+fn tkf91_indel_set_param_lambda() {
+    // arrange
+    let mut tkf91_indel_model =
+        TKF91IndelCostBuilder::new(1.0, 2.0, setup_test_phylo(dna_alphabet()))
+            .build()
+            .unwrap()
+            .model;
+
+    // act & assert
+    tkf_set_lambda(&mut tkf91_indel_model.clone());
```
junniest: I am confused, why do you need a clone here?

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+    let tkf_indel_model = TKF91IndelCostBuilder::new(1.1, 2.0, setup_test_phylo(dna_alphabet()))
+        .build()
+        .unwrap()
+        .model;
+
+    // act
+    let fmt = format!("{}", tkf_indel_model);
+
+    // assert
+    assert_eq!(fmt, "TKF91 with lambda = 1.1, mu = 2");
```
junniest: Here and below, I find that the comments only make the test longer and do not add to the meaning.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+    validate_r(0.0, DEFAULT_R);
+    validate_r(1.0, DEFAULT_R);
+    validate_r(1.5, DEFAULT_R);
+    validate_r(0.1, 0.1);
+}
+
+#[test]
+fn tkf91_model_fmt() {
+    // arrange
+    let tkf_indel_model = TKF91IndelCostBuilder::new(1.1, 2.0, setup_test_phylo(dna_alphabet()))
```
junniest: Why do you make the Cost here when you only need the model itself? I would suggest creating the model without the cost to test specific output.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,983 @@
...
+    let logl = tkf91_cost.logl();
+    let half_manual = tkf91_indel_logl_without_aggregation(&tkf91_cost.model, &tkf91_cost.phylo);
+    let mut manual_calculation = 0.0;
+    manual_calculation += (1.0 - lambda / mu).ln();
+    // immortal links
+    manual_calculation += log_i1(lambda, b(lambda, mu, tree.by_id("A1").blen));
+    manual_calculation += log_i1(lambda, b(lambda, mu, tree.by_id("B2").blen));
+    manual_calculation += log_i1(lambda, b(lambda, mu, tree.by_id("I3").blen));
+    manual_calculation += log_i1(lambda, b(lambda, mu, tree.by_id("C4").blen));
+    // first block
```
junniest: The block comments are helpful, but I'd suggest adding the coordinates of each block to make the computation connect better.

phylo/src/tkf_model/tkf91.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /// Sets the parameter if it is valid then returns true, otherwise the parameter is not changed
+    /// and false is returned.
+    /// This assumes that the other parameter is valid
+    fn set_param(&mut self, idx: usize, value: f64) -> bool {
+        let param = TKF91Parameters::from_primitive(idx);
+        match param {
+            TKF91Parameters::Lambda => {
+                if value > 0.0 && value < self.mu() {
```
junniest: I think that here you should use the `self.param_range()` instead of checking manually, just to make sure that the ranges checked are always the same. 

However, I also wonder if it makes sense to check the ranges every time a parameter is set if the ModelOptimiser will always select a parameter in the valid range. A user may set parameters incorrectly, but they could, e.g. first set lambda and then mu, which in combination are valid but are not valid when set in this order. We cannot always save users from themselves, and I don't know if we should try.

phylo/src/tkf_model/tkf91.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    pub fn build(self) -> Result<TKFIndelCost<AA, TKF91IndelModel>> {
+        let (lambda, mu) = validate_lambda_and_mu(self.lambda, self.mu);
+        let model = TKF91IndelModel {
+            params: vec![lambda, mu],
+        };
+        let info = TKFIndelModelInfo::new::<_, TKF91IndelModel>(&self.phylo);
+        Ok(TKFIndelCost {
+            model,
+            phylo: self.phylo.clone(),
```
junniest: Hm, do we need a clone here? Since `self` is consumed, it should be ok to leave out the clone (although Rust may complain, I'm not sure)

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+                        .set(node_id, true);
+                    insertion_prob
+                }
+            }
+            Action::Nothing => 1.0,
+        }
+    }
+}
+
+impl<AA: AncestralAlignment, T: TKFModel> ModelSearchCost for TKFIndelCost<AA, T> {
```
junniest: I think it might make sense to move this to a separate file as well, so that all generic TKF things are all together in the main mod file and specific models (TKF91, TKF92 and TKFIndels) are separate.

phylo/src/tkf_model/mod.rs
```diff
@@ -0,0 +1,538 @@
...
+    fn param_range(&self, idx: usize) -> ParamRange;
+    fn insertion_prob_at_root(&self) -> f64;
+    fn insertion_prob_at_non_root(&self, beta: f64) -> f64;
+    fn block_prob(&self, x: f64, block_len: usize) -> f64;
+    fn get_blocks<AA: AncestralAlignment>(msa: &AA) -> Vec<usize>;
+}
+
+// TODO: link our paper once it is published. For now see original TKF92 paper: https://doi.org/10.1007/bf00163848
+#[derive(Clone, Debug)]
+struct TKFIndelModelInfo {
```
junniest: Doesn't this apply to all TKF-style models? If so, the name is confusing because it seems like it is just meant for the TKF indel model without the substitution models, rather than all of them.

phylo/src/tkf_model/tkf92.rs
```diff
@@ -0,0 +1,287 @@
...
+                    self.one_minus_r_over_r = (1.0 - value) / value;
+                    return true;
+                }
+            }
+            TKF92Parameters::Invalid(_) => return false,
+        };
+        false
+    }
+
+    fn param_range(&self, idx: usize) -> crate::likelihood::ParamRange {
```
junniest: nitpick: unnecessary `crate::likelihood`

junniest (commented):
I have one major comment to the design at the moment. You have a lot of methods called `get_something` which to me indicates that they should only perform a computation and return the result. However, most if not all of these methods also set values in the temporary storage, meaning that they produce side effects.
My gut feeling is that methods with names like `get` should not produce side effects (set values) and methods with names like `set` should not return values, because otherwise it is very difficult to track where changes to the internal states are happening.

phylo/src/tkf_model/mod.rs
MattesMrzik: By adding this line, I wanted to initiate this discussion, because I wasn't sure if we really want this. What do you think?

phylo/src/tkf_model/mod.rs
MattesMrzik: oki, i moved this set to another place, the method now only determines the event (renamed from action)

phylo/src/tkf_model/mod.rs
MattesMrzik: Since i am avoiding caching for now. This mixing of returning a value vs setting sth is no longer an issue here.

phylo/src/tkf_model/mod.rs
MattesMrzik: removed caching which also removed this call

phylo/src/tkf_model/tests.rs
MattesMrzik: no

phylo/src/tkf_model/tests.rs
MattesMrzik: removed them

phylo/src/tkf_model/tests.rs
MattesMrzik: thanks. did so now in the private tests.

phylo/src/tkf_model/tkf91.rs
MattesMrzik: Oki, i removed the checks

phylo/src/tkf_model/mod.rs
MattesMrzik: Since this model info only cares about the indel process, i feel like the name is fitting. if we have tkf cost with substitutions, it's just tkf indel cost (with tkf indel model info) and substitution cost (with its own model info). Also, for the TKF with substitutions we simply do 
```rust
self.indel_cost.cost() + self.subst_cost.cost()
```
So both the indel cost and subst cost have their own model info

phylo/src/tkf_model/tests.rs
MattesMrzik: I removed this test since the fn no longer returns a bool

phylo/src/tkf_model/tests.rs
MattesMrzik: I removed this test since the fn no longer returns a bool

phylo/src/tkf_model/mod.rs
MattesMrzik: I added some comments. (see coming commit)

phylo/src/tkf_model/mod.rs
MattesMrzik: thanks for that suggestion. i implemented it.

phylo/src/tkf_model/mod.rs
junniest: Fair point!

phylo/src/tkf_model/mod.rs
MattesMrzik: I added a comment 
```
 // Since nothing happened, the previous event status remains the same.
```
This can be due to an insertion in another subtree, so for determining the previous event we must look back further. And we can look back further by not changing the previous event in that case.

phylo/src/tkf_model/mod.rs
junniest: That's a good question actually. I found this link: https://matklad.github.io/2021/07/09/inline-in-rust.html.
I think the bottom line is that we could add the`#[inline]` to the small generic functions here especially since they are not for public use, but it might be unnecessary because the compiler will do it anyway.
What do you think?

phylo/src/tkf_model/mod.rs
MattesMrzik: see #47

phylo/src/tkf_model/mod.rs
MattesMrzik: i changed the names, pls see next commit.

phylo/src/tkf_model/mod.rs
MattesMrzik: yes i meant eta. i did some renaming, pls see the next commit

phylo/src/tkf_model/mod.rs
MattesMrzik: i added some comments and did some renaming, pls see the next commit.

phylo/src/tkf_model/tests.rs
MattesMrzik: i added 
```

// This is a direct implementation of the TKF92 log likelihood calculation without any
// aggregation over subtrees and without substitutions. This direct calculation is sufficient
// if one is only interested in the indel likelihood for a fixed alignment and tree.
// Used for testing purposes only, i.e., to validate the aggregated implementation.
```

phylo/src/tkf_model/mod.rs
MattesMrzik: I added a comment: 
```rust
/// This struct holds intermediate values for the computation of the log likelihood
/// of an ancestral alignment and tree under a TKF indel model, i.e., without substitutions.
/// The intermediate values are needed for re-alignment.
```

phylo/src/tkf_model/mod.rs
MattesMrzik: moved it to a separate file

phylo/src/tkf_model/mod.rs
MattesMrzik: yep was also thinking that the compiler would probably do that. But perhaps only for n0 (which is just a multiplication) but not for eta. I would keep the #[inline] for these couple of easy funtion, which are not called in many places.

phylo/src/alignment/mod.rs
```diff
@@ -412,7 +414,7 @@ impl Alignment for MASA {
     /// assert_eq!(i1_seq, "XXX");
     /// let i1_map = phylo_info.msa.ancestral_map(&phylo_info.tree.by_id("I1").idx);
     /// assert_eq!(i1_map, &vec![Some(0), Some(1), None, Some(2)]);
-    /// /// or use the align_seq marco to test seq and map at the same time
+    /// /// or use the align_seq macro to test seq and map at the same time
```
junniest: nitpick: this should have `//` instead of `///` since the latter is just for docstring comments

phylo/src/alignment/mod.rs
```diff
@@ -446,6 +448,20 @@ impl AncestralAlignment for MASA {
...
+        &self.ancestral_maps
+    }
+
+    // This is needed because with the TKF models we need to re-estimate the ancestral maps after
+    // a tree move is applied.
+    fn update_ancestral_map(&mut self, node_idx: &NodeIdx, map: Mapping) {
+        if let Some(anc_map) = self.ancestral_maps.get_mut(node_idx) {
+            *anc_map = map;
+        } else {
+            panic!("NodeIdx {node_idx} is not an internal node");
```
junniest, line 461: Just wondering -- should this really panic? Or could you print a warning/error instead.
The reason why I'm asking is whether this would be something that breaks the reconstruction or could it just be safely ignored.

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,837 @@
...
+    let logl = tkf91_cost.logl();
+    let half_manual = tkf91_indel_logl_without_aggregation(&tkf91_cost.model, &tkf91_cost.phylo);
+    let mut manual_calculation = 0.0;
+    manual_calculation += (1.0 - lambda / mu).ln();
+    // immortal links
+    manual_calculation += log_i1(lambda, beta(lambda, mu, tree.by_id("A1").blen));
+    manual_calculation += log_i1(lambda, beta(lambda, mu, tree.by_id("B2").blen));
+    manual_calculation += log_i1(lambda, beta(lambda, mu, tree.by_id("I3").blen));
+    manual_calculation += log_i1(lambda, beta(lambda, mu, tree.by_id("C4").blen));
+    // first block ([0:2], insertion at C4)
```
junniest, line 493: That's perfect, thank you for adding the block explanation comments!

phylo/src/tkf_model/tests.rs
```diff
@@ -0,0 +1,837 @@
...
+        .build()
+        .unwrap();
+
+    // act & assert
+    // sanity check
+    let logl = ModelSearchCost::cost(&tkf_cost);
+    assert_eq!(logl, ModelSearchCost::cost(&tkf_cost));
+
+    // The likelihood should change if we change model parameters
+    match model_to_change {
```
junniest: I think it might be helpful here to split this huge template test function into two, one that changes the indel model and one that changes substitution model parameters. You wouldn't need this match then (which happens twice) and I think it would make the tests cleaner as well.

phylo/src/tkf_model/tkf92.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+
+/// Validates the TKF92 parameter r. If it is not valid, it is set to
+/// default value and a warning is logged.
+/// Returns valid r.
+fn validate_r(r: f64) -> f64 {
+    let mut valid_r = r;
+    if r == 0.0 {
+        warn!(
```
junniest: Nitpick: I think we've changed it somewhere else before, but it could be helpful to do the assignment first and then output, i.e.:
```
valid_r = DEFAULT_R;
warn!("Tried to set r to invalid value {r}. It must be in (0, 1). Setting r to {valid_r}");
```

phylo/src/tkf_model/tkf92.rs
```diff
@@ -0,0 +1,287 @@
...
+            params: vec![0.5, 1.0, 0.3],
+            log_r: 0.0,              // dummy
+            one_minus_r_over_r: 0.0, // dummy
+        };
+        // Use an invalid index
+        model.param_range(3);
+    }
+
+    #[test]
+    fn tkf92_model_fmt() {
```
junniest, line 257: Oh this test confused me for some reason, I was expecting the output to have 0.0 because of the dummy values that never get computed properly, but it's the parameters that are being printed... I guess this is perfectly fine in private tests, since the struct cannot be created by an external crate anyway. It's just always worrying if a struct can be created with invalid parameter values, in case it will get created like that and the rest of the code assumes it is set up correctly.
But in any case, in prod code it is only created with a builder, right?

phylo/src/tkf_model/tkf_indel.rs
```diff
@@ -0,0 +1,514 @@
...
+    /// The right exclusive interval borders of the blocks.
+    /// See [`TKFModel::get_blocks`].
+    blocks: Vec<usize>,
+    /// The lengths of the blocks.
+    /// See [`get_block_lengths`].
+    block_lengths: Vec<usize>,
+
+    /// previous_event_deletion[node] = true if the last event was a deletion for a that <node>.
+    /// See [`TKFIndelCost::determine_event`] and [`TKFIndelCost::update_previous_event`].
+    previous_event_deletion: Vec<bool>,
```
junniest: nitpick: could also be a `FixedBitSet`? We could make a newtype for this if it would help readability, but I think it would be ok anyway.

phylo/src/tkf_model/tkf_indel.rs
```diff
@@ -0,0 +1,514 @@
...
+    /// See [`TKFIndelCost::eta_for_non_root`].
+    node_eta: DMatrix<f64>,
+    /// subtree_eta[(node, block)] = sum of node_eta for all nodes in the subtree rooted in <node>
+    /// for the block with id <block>. Since we only have one insertion per column, at most one
+    /// node in the subtree can contribute to this sum.
+    subtree_eta: DMatrix<f64>,
+
+    /// beta[node] = beta(node.blen)), precomputed for each node.
+    /// See [`beta`] function.
+    beta: Vec<f64>,
```
junniest, line 85: nitpick: this and the ones below could also be `DVector` just for consistency? I am not sure if you ever need them as full vectors for any linear algebra, this is why I use those.

junniest: Update: Lol, I made chatGPT explain -- https://chatgpt.com/share/691b13dd-a1ac-800f-87fc-470837f5a34a.
Shouldn't make a difference and a regular vector is better, unless you use some vectorised operations on the vector (compute all values at the same time).

phylo/src/tkf_model/tkf_indel.rs
junniest: Heh, per chatGPT `FixedBitSet` pretty much always wins in terms of access times, so probably better use that, although it also shouldn't make much of a difference.

phylo/src/tkf_model/tkf_indel.rs
```diff
@@ -0,0 +1,514 @@
...
+            self.model_info.borrow().eta[usize::from(node_idx)]
+        } else {
+            0.0
+        }
+    }
+
+    fn event_prob_for_non_root(&self, node_idx: &NodeIdx, action: Event) -> f64 {
+        let node_id = usize::from(node_idx);
+        match action {
+            Event::Deletion => self.model_info.borrow_mut().n0[node_id],
```
junniest: nitpick: this doesn't need a `borrow_mut()`

junniest (approved):
There are some tiny nitpicks, but this looks great now! Well done! It is really well-decoupled now and it's a lot easier (at least for me) to understand what's going on.

phylo/src/alignment/mod.rs
MattesMrzik: I feel like panic is fitting since if this method is called, there is something wrong with the logic of the calling code. When people use `update_ancestral_map` and then have code where a leaf is passed, sth is flawed with their logic and its nice to point that out directly as one might miss the warning in the log.

phylo/src/tkf_model/tkf92.rs
MattesMrzik: yes it can only be created with a builder outside the private context. And as these are just some helper/cache vars they dont get printed.
i added a comment 
```rust
log_r: 0.0,              // cache filled with dummy since it is not printed
one_minus_r_over_r: 0.0, // cache filled with dummy since it is not printed

```

junniest: I think it's ok, thank you for clarifying the comment! I don't think this was a breaking issue, just got confused and thought it might be helpful to raise that (:

phylo/src/alignment/mod.rs
junniest: That's fair, I agree with this explanation!

PR #109, opened 2025-10-03
Fix/propagate optimisation conditions

junniest:
Propagating optimisation conditions from initial settings to optimiser setup (e.g. epsilon)

### DIFF
--- a/phylo/src/optimisers/model_optimiser.rs
+++ b/phylo/src/optimisers/model_optimiser.rs
@@ -165,3 +165,84 @@ impl<C: ModelSearchCost> CostFunction for ParamOptimiser<C> {
         true
     }
 }
+
+#[cfg(test)]
+#[cfg_attr(coverage, coverage(off))]
+mod tests {
+    use std::path::Path;
+
+    use assert_matches::assert_matches;
+
+    use crate::likelihood::ModelSearchCost;
+    use crate::phylo_info::PhyloInfoBuilder as PIB;
+    use crate::pip_model::{PIPCostBuilder as PIPCB, PIPModel};
+    use crate::substitution_models::{
+        dna_models::GTR, protein_models::WAG, SubstModel, SubstitutionCostBuilder as SCB,
+    };
+
+    use super::*;
+
+    #[test]
+    fn estimated_freqs_default_to_empirical_dna() {
+        let fldr = Path::new("./data/sim");
+        let info = PIB::with_attrs(fldr.join("K80/K80.fasta"), fldr.join("tree.newick"))
+            .build()
+            .unwrap();
+
+        let model = SubstModel::<GTR>::new(&[], &[]);
+        let cost = SCB::new(model, info).build().unwrap();
+        let mut opt = ModelOptimiser::new(cost.clone(), FrequencyOptimisation::Estimated);
+        assert_matches!(opt.freq_opt, FrequencyOptimisation::Estimated);
+
+        opt.optimise_frequencies();
+        assert_eq!(opt.c.freqs(), &opt.c.empirical_freqs());
+    }
+
+    #[test]
+    fn estimated_freqs_default_to_empirical_aa() {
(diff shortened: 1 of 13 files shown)


phylo/src/optimisers/model_optimiser.rs
```diff
@@ -7,86 +7,116 @@ use log::{debug, info, warn};
...
     pub fn run(mut self) -> Result<ModelOptimisationResult<C>> {
         info!("Optimising the evolutionary model: {}", self.c);
+        info!("Optimisation stopping condition: {}", self.stop_condition);
 
         let init_cost = self.c.cost();
         info!("Initial cost: {init_cost}");
-        let mut curr_cost = init_cost;
+
+        let mut curr_cost = self.optimise_frequencies();
+        debug_assert!(curr_cost >= init_cost);
```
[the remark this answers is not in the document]
MattesMrzik: This comment is wrong since we do 
```rust
let init_cost = self.c.cost();
curr_cost = self.optimise_frequencies();
let mut costs = vec![init_cost, curr_cost];
```

phylo/src/optimisers/model_optimiser.rs
```diff
@@ -134,3 +165,84 @@ impl<C: ModelSearchCost> CostFunction for ParamOptimiser<C> {
...
+    #[test]
+    fn estimated_freqs_default_to_empirical_dna() {
+        let fldr = Path::new("./data/sim");
+        let info = PIB::with_attrs(fldr.join("K80/K80.fasta"), fldr.join("tree.newick"))
+            .build()
+            .unwrap();
+
+        let model = SubstModel::<GTR>::new(&[], &[]);
+        let cost = SCB::new(model, info).build().unwrap();
+        let mut opt = ModelOptimiser::new(cost.clone(), FrequencyOptimisation::Estimated);
```
MattesMrzik, line 194: Why not use `FrequencyOptimisation::Empirical` here? The tests are called ...empirical... and you assert against `&opt.c.empirical_freqs()`.
(See also the tests below)

phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -481,55 +481,60 @@ fn pip_optimise_model_tree() {
...
+        model_tree_opt
+            .cost
+            .tree()
+            .robinson_foulds(tree_opt.cost.tree())
+            <= 2
     );
+    assert!(model_tree_opt.final_cost > tree_opt.final_cost);
+
+    // Check that the optimised tree+model is better than the starting tree with the optimised model
+    let old_tree_optimised_model_logl = PIPCB::new(optimised_pip, tree_opt.cost.info.clone())
```
MattesMrzik: How is this the old tree if you use `tree_opt` which already did optimize the tree? Is it because you run the topo opti twice and you are checking the final tree vs the second tree?

MattesMrzik (approved):
Good PR

phylo/src/optimisers/model_optimiser.rs
junniest: These are actually checking that the `FrequencyOptimisation::Estimated` variant ends up defaulting to empirical frequencies, since proper ML estimation is not actually implemented. These will get removed once the estimation is implemented properly, it was just to make sure that the behaviour is meaningful! The name of this test is actually `estimated_freqs_default_to_empirical_dna`, apologies for the confusion!

phylo/src/optimisers/topo_optimiser_tests.rs
junniest: This is indeed confusing... What I tried to say is that the `old_tree` is the tree optimised without model optimisation and if we use the optimised model to compute the likelihood it should still be better. I will rename this to `tree_opt` to match the naming scheme in the rest of the test. I don't think this is a good naming scheme anyway, but that was the best I could come up with to try and distinguish between different objects.

PR #111, opened 2025-10-10
Enhancement/cost cache valid tests

junniest:
Added a bunch of tests for `PIPCost` and `SubstitutionCost` to make sure that the costs always match when the tree gets dirty, a branch length changes, or model parameters or frequencies change.

### DIFF
--- a/phylo/src/pip_model/tests.rs
+++ b/phylo/src/pip_model/tests.rs
@@ -858,9 +858,7 @@ fn blen_leading_to_minusinf() {
 }
 
 #[cfg(test)]
-fn dirty_tree_costs_match_template<Q: QMatrix + QMatrixMaker>(alphabet: Alphabet) {
-    use crate::likelihood::TreeSearchCost;
-
+fn setup_test_info(alphabet: Alphabet) -> PhyloInfo<MSA> {
     let tree = tree!("(((A:1.0,B:1.0)E:2.0,(C:1.0,D:1.0)F:2.0)G:3.0);");
     let msa = MSA::from_aligned(
         Sequences::with_alphabet(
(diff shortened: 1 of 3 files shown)


phylo/src/pip_model/tests.rs
```diff
@@ -854,3 +856,222 @@ fn blen_leading_to_minusinf() {
...
+        ),
+        &tree,
+    )
+    .unwrap();
+    let info = PhyloInfo { msa, tree };
+
+    let model = PIPModel::<Q>::new(&[], &[]);
+    let mut c = PIPB::new(model.clone(), info.clone()).build().unwrap();
+    let logl = TreeSearchCost::cost(&c);
+    assert_eq!(logl, TreeSearchCost::cost(&c));
```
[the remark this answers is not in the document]
MattesMrzik: The cost has an internal refcell so change is not impossible. (This is a remark on a comment copilot made)

phylo/src/pip_model/tests.rs
```diff
@@ -854,3 +856,177 @@ fn blen_leading_to_minusinf() {
...
+
+    // The likelihood should be the same if we make the tree dirty without changing it
+    let mut tree = c.info.tree.clone();
+    tree.dirty();
+    c.update_tree(tree);
+    assert_eq!(logl, TreeSearchCost::cost(&c));
```
MattesMrzik, line 894: why are you using TreeSearchCost::cost(&c) and not c.cost()

MattesMrzik (approved):
Looks good

phylo/src/pip_model/tests.rs
junniest: This is because the ModelSearchCost trait also has a cost method, and the compiler is not sure (rightfully) which one to pick. This notation clearly states that it should use the TreeSearchCost trait's method, not the ModelSearchCost's method (even though in this case they do the same thing). There are a few tests doing this too, see `likelihood/tests.rs:18`.

PR #114, opened 2025-10-15
87 readme to tests

### LINKED ISSUE #87: Add example code from README.md to tests
Currently, changes that break the example `main()` from the README.md need to be recognized manually. We should add a test function that is identical to the block of the `main()` and add comments to it saying: "If you change this code, then pls also update the example main() in the README.md". And also the other way around, i.e. mention in the README.md example `main()`  to also keep the test up-to-date. For this we could use sth like this
```markdown
[comment]: # (comment text)
```
or
```markdown
[//]: # (This is a hidden comment that will not show on GitHub)

```
since this will be visible when looking at the raw file, but these comments are not visible on the rendered markdown preview.

MattesMrzik:
closes #87 by adding comments to both the README.md and test file, prompting contributors to keep the example from the README.md and the test in sync.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,6 @@ MSRV detected using [`cargo-msrv`]( https://github.com/foresterre/cargo-msrv ).
 <!-- WARNING: If you change the code below, please also update the test "phylo/src/optimisers/topo_optimiser_tests.rs:example_main_from_readme" accordingly!) -->
 
 ```rust
-use std::path::Path;
-
 use phylo::likelihood::TreeSearchCost;
 use phylo::optimisers::TopologyOptimiser;
 use phylo::phylo_info::PhyloInfoBuilder;
--- a/README.md
+++ b/README.md
@@ -54,7 +54,6 @@ fn main() -> std::result::Result<(), anyhow::Error> {
     let c = SubstitutionCostBuilder::new(k80, info).build()?;
     let unopt_cost = c.cost();
     let rng = FakeGenerator::default();
-
     let optimiser = TopologyOptimiser::new(c, SprOptimiser {}, &rng);
     let result = optimiser.run()?;
     assert_eq!(unopt_cost, result.initial_cost);
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ MSRV detected using [`cargo-msrv`]( https://github.com/foresterre/cargo-msrv ).
 
 ### Example
 
-[comment]: # (WARNING: If you change the code below, please also update the test "example_main_from_readme" accordingly!)
+<!-- WARNING: If you change the code below, please also update the test "phylo/src/optimisers/topo_optimiser_tests.rs:example_main_from_readme" accordingly!) -->
 
 ```rust
 use std::path::Path;
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ use phylo::substitution_models::{SubstModel, SubstitutionCostBuilder, K80};
 
 fn main() -> std::result::Result<(), anyhow::Error> {
     // Note: This example uses test data from the repository
-    let info = PhyloInfoBuilder::new(Path::new("./examples/data/K80.fasta")).build()?;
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -37,21 +37,24 @@ MSRV detected using [`cargo-msrv`]( https://github.com/foresterre/cargo-msrv ).
 
 ### Example
 
+[comment]: # (WARNING: If you change the code below, please also update the test "example_main_from_readme" accordingly!)
```
junniest: hmmm, this is interesting, I've never seen markdown comments like this. Ah, it's probably because it's not a real comment (; Judging by this: https://stackoverflow.com/questions/4823468/comments-in-markdown, it's abuse of link labels that would remove the comment from rendered HTML to hide it completely. In this case I don't see why we should use this because we don't really need to hide it. I would suggest using regular old-school html comment as elsewhere in readme:

This would also remove the random # and the parentheses.

junniest: Could you also provide the full path to the test? I guess it's `phylo/src/optimisers/topo_optimiser_tests.rs:example_main_from_readme`. Otherwise once this change is decoupled from the source file it will be very difficult to find (or rather seems illogical).
Another option could be to move this one into library integration tests altogether because then there would be no need for the warnings in the test itself (or into examples, which might be appropriate in this case).

README.md
```diff
@@ -37,21 +37,24 @@ MSRV detected using [`cargo-msrv`]( https://github.com/foresterre/cargo-msrv ).
...
+
 use phylo::likelihood::TreeSearchCost;
 use phylo::optimisers::TopologyOptimiser;
 use phylo::phylo_info::PhyloInfoBuilder;
 use phylo::substitution_models::{SubstModel, SubstitutionCostBuilder, K80};
 
 fn main() -> std::result::Result<(), anyhow::Error> {
     // Note: This example uses test data from the repository
-    let info = PhyloInfoBuilder::new(Path::new("./examples/data/K80.fasta").to_path_buf()).build()?;
+    let info = PhyloInfoBuilder::new(Path::new("./examples/data/K80.fasta")).build()?;
```
junniest: Actually this doesn't even have to have Path::new() anymore, I changed the signature of `PhyloInfoBuilder` to `new(sequence_file: impl AsRef<Path>)` a while ago, but likely forgot to change the example! Just tells you how difficult it is to maintain such examples in practice...

So this line can be:
`let info = PhyloInfoBuilder::new("./examples/data/K80.fasta").build()?;`
and we can remove the `use std::path::Path;`.

phylo/src/optimisers/topo_optimiser_tests.rs
```diff
@@ -879,3 +879,33 @@ fn max_iter() {
     assert!(result.final_cost >= res_default.final_cost);
     assert_eq!(result.initial_cost, unopt_cost);
 }
+
+// Warning: If you change this test, please also update the example in the README.md file!
+// Additionally: use "use crate::..." in this test here, but "use phylo::..." in the README.md
```
junniest: I think that unfortunately after the first warning this additional warning somehow loses it's meaning (doesn't seem as serious). I think we should move this to examples/integration tests, which would remove the need for tor this change. The more changes we need to propagate, the higher the chance of error.

junniest (approved):
Thanks for taking care of this and cleaning up the Readme example! I added some comments on how to make us less prone to forget to update either place.

README.md
MattesMrzik: yep, but we dont have integration tests yet, right?

phylo/src/optimisers/topo_optimiser_tests.rs
MattesMrzik: Adjusted it to 
```
// Warning: If you change this test, please also update the example in the README.md file
// accordingly!
```

README.md
MattesMrzik: Please see #118

junniest (approved):
LGTM! Merge ahead!

PR #117, opened 2025-10-22
Implemented valid param ranges for ModelSearchCost

### LINKED ISSUE #116: Valid ranges for parameter optimisation
Parameters are optimized using Brent, which takes a range (lower, upper limit). Currently, the dynamic range (epsilon, start value for parameter * 100) is used. However, some parameters are only valid in a particular range (like the r in TKF92 which must be in (0,1)). Therefore, we need to add a method to the trait `ModelSearchCost`, which returns for each parameter its valid range.

MattesMrzik:
Closes: #116

### DIFF
--- a/phylo/src/likelihood/mod.rs
+++ b/phylo/src/likelihood/mod.rs
@@ -1,7 +1,7 @@
 use crate::substitution_models::FreqVector;
 use crate::tree::Tree;
 
-/// The valid range for a model parameter \[min, max], inclusive.
+/// The valid range for a model parameter [min, max], inclusive.
 pub type ParamRange = (f64, f64);
 /// Parameter is unbounded (i.e. can take any real value).
 pub static PARAM_RANGE_UNBOUNDED: ParamRange = (f64::MIN, f64::MAX);
@@ -9,7 +9,7 @@ pub static PARAM_RANGE_UNBOUNDED: ParamRange = (f64::MIN, f64::MAX);
 pub static PARAM_RANGE_POSITIVE: ParamRange = (f64::EPSILON, f64::MAX);
 /// Parameter is non-negative (i.e. greater than or equal to zero).
 pub static PARAM_RANGE_NON_NEGATIVE: ParamRange = (0.0, f64::MAX);
-/// Parameter is in the unit interval \[0, 1], inclusive.
+/// Parameter is in the unit interval [0, 1], inclusive.
 pub static PARAM_RANGE_UNIT_INTERVAL: ParamRange = (0.0, 1.0);
 /// Parameter is in the unit interval (0, 1), exclusive.
 pub static PARAM_RANGE_UNIT_INTERVAL_EXCLUSIVE: ParamRange = (f64::EPSILON, 1.0 - f64::EPSILON);
@@ -35,10 +35,10 @@ pub trait ModelSearchCost {
     // The optimisers will maximise the cost, so if the cost should be minimised instead, it should be negated.
     // The likelihood or the log-likelihood are maximised, the parsimony score is minimised.
     fn cost(&self) -> f64;
+    fn param_count(&self) -> usize;
     fn param(&self, param: usize) -> f64;
     fn set_param(&mut self, param: usize, value: f64);
-    fn param_count(&self) -> usize;
-    /// Returns the valid range for the given parameter (index), interval is inclusive.
+    /// Returns the valid range for a model parameter [min, max], inclusive.
     fn param_range(&self, param: usize) -> ParamRange;
     fn set_freqs(&mut self, freqs: FreqVector);
     fn empirical_freqs(&self) -> FreqVector;
--- a/phylo/src/likelihood/mod.rs
+++ b/phylo/src/likelihood/mod.rs
@@ -1,6 +1,8 @@
 use crate::substitution_models::FreqVector;
 use crate::tree::Tree;
 
+pub type ParamRange = (f64, f64);
+
 pub trait TreeSearchCost {
     // The optimisers will maximise the cost, so if the cost should be minimised instead, it should be negated.
     // The likelihood or the log-likelihood are maximised, the parsimony score is minimised.
@@ -20,8 +22,11 @@ pub trait ModelSearchCost {
     // The optimisers will maximise the cost, so if the cost should be minimised instead, it should be negated.
     // The likelihood or the log-likelihood are maximised, the parsimony score is minimised.
     fn cost(&self) -> f64;
+    fn param(&self, param: usize) -> f64;
     fn set_param(&mut self, param: usize, value: f64);
-    fn params(&self) -> &[f64];
+    fn param_count(&self) -> usize;
+    /// Returns the valid range for the given parameter (index), interval is inclusive.
+    fn param_range(&self, param: usize) -> ParamRange;
     fn set_freqs(&mut self, freqs: FreqVector);
     fn empirical_freqs(&self) -> FreqVector;
     fn freqs(&self) -> &FreqVector;
--- a/phylo/src/likelihood/mod.rs
+++ b/phylo/src/likelihood/mod.rs
@@ -1,7 +1,20 @@
 use crate::substitution_models::FreqVector;
 use crate::tree::Tree;
 
+/// The valid range for a model parameter \[min, max], inclusive.
 pub type ParamRange = (f64, f64);
+/// Parameter is unbounded (i.e. can take any real value).
+pub static PARAM_RANGE_UNBOUNDED: ParamRange = (f64::MIN, f64::MAX);
+/// Parameter is strictly positive (i.e. greater than zero).
+pub static PARAM_RANGE_POSITIVE: ParamRange = (f64::EPSILON, f64::MAX);
+/// Parameter is non-negative (i.e. greater than or equal to zero).
+pub static PARAM_RANGE_NON_NEGATIVE: ParamRange = (0.0, f64::MAX);
+/// Parameter is in the unit interval \[0, 1], inclusive.
+pub static PARAM_RANGE_UNIT_INTERVAL: ParamRange = (0.0, 1.0);
+/// Parameter is in the unit interval (0, 1), exclusive.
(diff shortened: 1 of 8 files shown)


MattesMrzik:
Some questions/remarks:
- Should I add tests for each model, checking if the correct ParamRange is returned?
- the method `params` of trait `QMatrix` is perhaps not really necessary. Keep it or ditch it?
- I sorted the fns of `impl QMatrix for JC69 {` and similar according to the trait def, that's why there are so many changed lines.

phylo/src/optimisers/model_optimiser.rs
```diff
@@ -128,8 +127,9 @@ impl<C: ModelSearchCost + Display + Clone> ModelOptimiser<C> {
             cost: RefCell::new(self.c.clone()),
             param,
         };
-        let min = f64::EPSILON;
-        let max = start_value * 100.0;
+        let range = self.c.param_range(param);
+        let min = range.0;
+        let max = range.1.min(start_value * 100.0);
```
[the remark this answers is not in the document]
MattesMrzik: Since the start value is always in the valid range, the max is always bigger than min. Add a (debug) assert or warning + set param to within valid range if this is not true??

phylo/src/optimisers/model_optimiser_tests.rs
```diff
@@ -28,13 +28,13 @@ fn likelihood_improves_k80() {
         .unwrap();
     assert!(o.final_cost > unopt_logl);
 
-    let model = SubstModel::<K80>::new(o.cost.freqs().into(), o.cost.params());
+    let model = SubstModel::<K80>::new(o.cost.freqs().into(), o.cost.model.params());
```
junniest: Ah damnit, now I see why the cost also had the `params()` method... This really gets a bit silly unfortunately, that all the 3 traits (`ModelSearchCost`, `EvoModel` and `QMatrix`) have the same set of methods (`param`, `set_param`, `param_count`), but I don't really see how to avoid that at the moment...

However, for these tests, just to make it a bit cleaner, I think we can remove recreating the model. We don't need to check that `SubstModel::new` does the right thing (this should be tested elsewhere anyway). So what we could do is this:

```
    let model = o.cost.model.clone();
    assert_relative_eq!(model.params()[0], 1.884815, epsilon = 1e-5);

    let c2 = SCB::new(model, info).build().unwrap();
    assert_relative_eq!(o.initial_cost, unopt_logl);
    assert_relative_eq!(o.final_cost, o.cost.cost());
    assert_relative_eq!(o.final_cost, c2.cost());
    assert_relative_eq!(o.final_cost, -4034.500803, epsilon = 1e-5);
```

I think this looks a bit cleaner like that, otherwise accessing the different parameters feels awkward as hell to me.

phylo/src/optimisers/model_optimiser_tests.rs
```diff
@@ -399,11 +399,11 @@ fn arpip_example() {
     assert_relative_eq!(o.initial_cost, -212.3260420492571, epsilon = 1e-6); // value from the python script
 
     assert!(o.final_cost > initial_logl);
-    assert_ne!(o.cost.params(), pip_gtr.params());
+    assert_ne!(o.cost.model.params(), pip_gtr.params());
     assert_ne!(o.cost.freqs(), pip_gtr.freqs());
     assert_eq!(o.final_cost, o.cost.cost());
 
-    let pip_gtr = PIPModel::<GTR>::new(o.cost.freqs().as_slice(), o.cost.params());
+    let pip_gtr = PIPModel::<GTR>::new(o.cost.freqs().as_slice(), o.cost.model.params());
```
junniest: I think, again, this new is unnecessary, the next line can just be:
```
    let c = PIPCostBuilder::new(o.cost.model, info.clone())
        .build()
        .unwrap();
```

phylo/src/optimisers/model_optimiser_tests.rs
```diff
@@ -481,7 +481,7 @@ fn pip_gtr_optimisation() {
     let pip_o = ModelOptimiser::new(c, FrequencyOptimisation::Fixed)
         .run()
         .unwrap();
-    let pip_gtr = PIPModel::<GTR>::new(pip_o.cost.freqs().as_slice(), pip_o.cost.params());
+    let pip_gtr = PIPModel::<GTR>::new(pip_o.cost.freqs().as_slice(), pip_o.cost.model.params());
 
     let c = PIPCostBuilder::new(pip_gtr, info.clone()).build().unwrap();
```
junniest: I honestly don't understand why I was so bent on remaking all the models lol. Should be this:
```
    let c = PIPCostBuilder::new(pip_o.cost.model.clone(), info.clone())
        .build()
        .unwrap();
```

phylo/src/optimisers/model_optimiser_tests.rs
```diff
@@ -536,8 +540,8 @@ fn pip_protein_example() {
         .unwrap();
     assert!(o.final_cost > initial_logl);
     assert_relative_eq!(o.initial_cost, initial_logl);
-    assert_ne!(o.cost.params()[0], 2.0);
-    assert_ne!(o.cost.params()[1], 0.1);
+    assert_ne!(o.cost.model.params()[0], 2.0);
+    assert_ne!(o.cost.model.params()[1], 0.1);
```
junniest: Hm, I think since we mostly care about the cost doing the right thing here (or even the model optimiser, but that works on the cost), I think these two checks should rather be:
```
    assert_ne!(o.cost.param(0), 2.0);
    assert_ne!(o.cost.param(1), 0.1);
```

phylo/src/optimisers/model_optimiser_tests.rs
```diff
@@ -562,8 +566,8 @@ fn stop_condition_epsilon() {
 
     assert!(result.final_cost > initial_logl);
     assert_relative_eq!(result.initial_cost, initial_logl);
-    assert_ne!(result.cost.params()[0], 2.0);
-    assert_ne!(result.cost.params()[1], 0.1);
+    assert_ne!(result.cost.model.params()[0], 2.0);
+    assert_ne!(result.cost.model.params()[1], 0.1);
```
junniest: Same here, I think it's best to do:
```
    assert_ne!(result.cost.param(0), 2.0);
    assert_ne!(result.cost.param(1), 0.1);
```

phylo/src/optimisers/model_optimiser_tests.rs
```diff
@@ -454,7 +454,7 @@ fn pip_vs_python_no_gaps() {
     let o = ModelOptimiser::new(c, FrequencyOptimisation::Fixed)
         .run()
         .unwrap();
-    let params = o.cost.params();
+    let params = o.cost.model.params();
```
junniest: This can go I think and the checks below should rather be:
```
    assert_ne!(o.cost.param(0), 1.2);
    assert_ne!(o.cost.param(1), 0.45);
    assert_ne!(o.cost.param(2), 1.0);
```

phylo/src/substitution_models/mod.rs
```diff
@@ -180,13 +188,21 @@ impl<Q: QMatrix, A: Alignment> ModelSearchCost for SubstitutionCost<Q, A> {
...
         self.tmp.borrow_mut().node_models_valid.fill(false);
     }
 
-    fn params(&self) -> &[f64] {
-        self.model.params()
+    fn param_count(&self) -> usize {
+        self.model.qmatrix.param_count()
+    }
+
+    fn param_range(&self, param: usize) -> crate::likelihood::ParamRange {
```
junniest: nitpick: unnecessary `crate::likelihood::`

phylo/src/likelihood/mod.rs
```diff
@@ -1,6 +1,21 @@
...
 
+/// The valid range for a model parameter \[min, max], inclusive.
+pub type ParamRange = (f64, f64);
+/// Parameter is unbounded (i.e. can take any real value).
+pub static PARAM_RANGE_UNBOUNDED: ParamRange = (f64::MIN, f64::MAX);
+/// Parameter is strictly positive (i.e. greater than zero).
+pub static PARAM_RANGE_POSITIVE: ParamRange = (f64::EPSILON, f64::MAX);
+/// Parameter is non-negative (i.e. greater than or equal to zero).
+pub static PARAM_RANGE_NON_NEGATIVE: ParamRange = (0.0, f64::MAX);
+/// Parameter is in the unit interval \[0, 1], inclusive.
```
junniest: nitpick: is the `\` necessary?

phylo/src/likelihood/mod.rs
```diff
@@ -20,8 +35,11 @@ pub trait ModelSearchCost {
     // The optimisers will maximise the cost, so if the cost should be minimised instead, it should be negated.
     // The likelihood or the log-likelihood are maximised, the parsimony score is minimised.
     fn cost(&self) -> f64;
+    fn param(&self, param: usize) -> f64;
     fn set_param(&mut self, param: usize, value: f64);
-    fn params(&self) -> &[f64];
+    fn param_count(&self) -> usize;
```
junniest: nitpick: if we're trying to have a logical ordering of methods, would it maybe make more sense to have the following order:
```
    fn param_count(&self) -> usize;
    fn param(&self, param: usize) -> f64;
    fn set_param(&mut self, param: usize, value: f64);
    fn param_range(&self, param: usize) -> ParamRange;
```
Since `param_count` refers to all parameters and the next 3 methods are per parameter.

phylo/src/optimisers/model_optimiser.rs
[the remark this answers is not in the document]
junniest: I think Copilot is being too much, I don't think this is something that should be handled here. This can operate under the assumption that the range is valid and min <= start_value <= max.

But I think that adding a debug_assert! will not hurt regardless, so adding something like this is a good idea:
```
debug_assert!(min <= start_value && start_value <= max);
```

junniest (approved):
Looks good!

phylo/src/likelihood/mod.rs
MattesMrzik: i have checked again, and the [ are generated correctly in the docs. My IDE however wasnt rendering them, that's why i added it, but will remove it.

Issue #90, opened 2025-09-03
Refactor randomisers


junniest:
Refactor misleading/confusing features of the current randomiser implementation, such as:
1. Potentially misleading struct/trait names (`RandomSource` exists in `rand`, `SeededRng` is too similar to `SeedableRng`);
2. The randomiser is not crypto-safe which is never stated explicitly, it is suitable for most comp bio, but that should be made clear;
3. The randomiser is not thread-reproducible (i.e. if individual threads would need random numbers computational reproducibility goes out the window). This should at least be mentioned;
4. PRNG should be replaced by a platform-independent one instead of `StdRng`;
5. Remove Mutex from RandomGenerator since it is not going to maintain reproducibility across threads anyways, pass `&mut self` to methods instead;
6. ...

junniest:
Closed with #120

PR #120, opened 2025-11-03
Enhancement/refactor randomisers

junniest:
Addressing #90 to remove traits/structs that have similar names to objects existing in rand already, removing unnecessary Mutex'es and simplifying the random number generators in general.

### DIFF
--- a/phylo/examples/rng_examples.rs
+++ b/phylo/examples/rng_examples.rs
@@ -1,4 +1,4 @@
-use rand::distributions::WeightedIndex;
+use rand::distr::weighted::WeightedIndex;
 use rand::rngs::{SmallRng, StdRng};
 use std::iter::repeat_n;
 
--- a/phylo/examples/rng_examples.rs
+++ b/phylo/examples/rng_examples.rs
@@ -6,7 +6,7 @@ fn main() {
     // 1. Using the default RNG (StdRng-based)
     let seed = 42;
     println!("1. RNG (StdRng) with seed {seed} - reproducible:");
-    let rng = DefaultGenerator::new(seed);
+    let mut rng = DefaultGenerator::new(seed);
     for i in 0..3 {
         println!("  {i}: {:.6}", rng.gen::<f64>());
     }
@@ -20,7 +20,7 @@ fn main() {
 
     // 2. Generating from default RNG:
     println!("2. Generating from default RNG:");
-    let rng = DefaultGenerator::default();
+    let mut rng = DefaultGenerator::default();
     println!("  Random f64: {:.6}", rng.gen::<f64>());
     println!("  Random u32: {}", rng.gen::<u32>());
     println!("  Random i32: {}", rng.gen::<i32>());
@@ -34,7 +34,7 @@ fn main() {
     // 3. Creating a custom StdRng instance
     let seed = 123;
     println!("3. Custom StdRng instance with seed {seed} - reproducible:");
-    let custom_std_rng = RandomGenerator::<StdRng>::new(seed);
+    let mut custom_std_rng = RandomGenerator::<StdRng>::new(seed);
     for i in 0..3 {
         println!("  {i}: {:.6}", custom_std_rng.gen::<f64>());
     }
@@ -63,7 +63,7 @@ fn main() {
     // 5. Using a different seed for another StdRng instance
     let seed = 456;
     println!("5. Another StdRng instance with different seed {seed}:");
-    let another_std_rng: RandomGenerator<StdRng> = RandomGenerator::new(seed);
+    let mut another_std_rng: RandomGenerator<StdRng> = RandomGenerator::new(seed);
     for i in 0..3 {
         println!("  {i}: {:.6}", another_std_rng.gen::<f64>());
     }
@@ -71,7 +71,7 @@ fn main() {
 
     // 6. Creating a custom RNG instance with a different generator
     println!("6. Custom RNG instance with SmallRng with seed {seed}:");
-    let secure_rng: RandomGenerator<SmallRng> = RandomGenerator::new(seed);
+    let mut secure_rng: RandomGenerator<SmallRng> = RandomGenerator::new(seed);
     for i in 0..3 {
         println!("  {i}: {:.6}", secure_rng.gen::<f64>());
     }
--- a/phylo/examples/rng_examples.rs
+++ b/phylo/examples/rng_examples.rs
@@ -1,28 +1,16 @@
 use rand::distr::weighted::WeightedIndex;
-use rand::rngs::{SmallRng, StdRng};
+use rand::rngs::StdRng;
 use std::iter::repeat_n;
 
 use phylo::random::{DefaultGenerator, RandomGenerator};
 
 fn main() {
-    // 1. Using the default RNG (StdRng-based)
-    let seed = 42;
-    let mut rng = DefaultGenerator::new(seed);
-    println!("1. RNG (StdRng) with seed {} - reproducible:", rng.seed());
-    for i in 0..3 {
-        println!("  {i}: {:.6}", rng.random::<f64>());
-    }
-    // Reset and show reproducibility
-    rng.reseed(seed);
-    println!("After resetting to seed {}:", rng.seed());
-    for i in 0..3 {
-        println!("  {i}: {:.6}", rng.random::<f64>());
-    }
-    println!();
-
-    // 2. Generating from default RNG:
-    println!("2. Generating from default RNG:");
+    // 1. Generating from default RNG (Pcg32, seeded with the current timestamp):
     let mut rng = DefaultGenerator::default();
+    println!(
+        "1. Generating from default RNG (Pcg32-based), seeded with the current timestamp {}:",
+        rng.seed()
+    );
     println!("  Random f64: {:.6}", rng.random::<f64>());
     println!("  Random u32: {}", rng.random::<u32>());
     println!("  Random i32: {}", rng.random::<i32>());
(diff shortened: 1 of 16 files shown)


junniest:
Refactored randomisers, changed the default random to ChaCha8Rng for platform independence.

phylo/src/random/fake_random.rs
```diff
@@ -1,477 +1,305 @@
...
+    fn fake_generator_defaults() {
+        // Test FakeGenerator defaults
+        let mut fake_rng = FakeGenerator::default();
+        assert_eq!(fake_rng.seed(), 0);
+        let val: u64 = fake_rng.random();
+        assert_eq!(val, 0);
+        let val: f64 = fake_rng.random();
+        assert_eq!(val, 0.0);
+        let val: bool = fake_rng.random();
+        assert!(!val);
```
MattesMrzik: Isnt this the same as `fake_rng_defaults` above?

phylo/src/random/fake_random.rs
```diff
@@ -1,477 +1,305 @@
...
-        fake_rng.reseed(0);
+    fn fake_rng_with_u64_values() {
+        // FakeRng with pre-configured values
+        let values = (15..25).collect::<Vec<u64>>();
+        let mut fake_rng = RandomGenerator::from_rng(FakeRng::from_u64_values(values.clone()));
+        assert_eq!(fake_rng.seed(), 0);
         for i in 0..10 {
-            assert_eq!(fake_rng.gen_bool(0.7), values[i % values.len()]);
+            let val: u64 = fake_rng.random();
+            assert_eq!(val, values[i % values.len()]);
```
MattesMrzik: perhaps loop over 0..30, because otherwise the `i % values.len()` is never wrapped around.

phylo/src/random/fake_random.rs
```diff
@@ -1,477 +1,305 @@
...
-        let val: f64 = fake_rng.gen();
-        assert_eq!(val, values[0]); // Wraps around
-        let val: u64 = fake_rng.gen();
-        assert_eq!(val, 0); // Default for u64
-        assert!(!fake_rng.gen::<bool>()); // Default for bool
+    fn fake_sample_default() {
+        // always returns first non-zero index when no pre-configured values
+        let mut rng = FakeGenerator::default();
+        let dist = WeightedIndex::new([1.0, 2.0, 3.0]).unwrap();
+        assert_eq!(rng.sample(&dist), 0);
```
MattesMrzik, line 247: Could we also make a similar test like 
```rust
let floats = vec![0.1, 0.25, 0.5, 0.75, 0.9];
let mut rng = RandomGenerator::from_rng(FakeRng::from_f64_values(floats));
let dist = WeightedIndex::new([1.0, 2.0, 3.0]).unwrap();
assert_eq!(rng.sample(&dist), ...);
```
and then assert to whatever value actually turns up?
just asking out of curiosity.

This might be useful to know in the tree builder. Since if we want to use this sampler to pick two nodes that should be joined next in the temperature version of the nj builder, it might be nicer to set up the fake rng to not always pick the first non-zero entry of the WeightedIndex.

phylo/src/random/mod.rs
```diff
@@ -1,198 +1,145 @@
...
 /// ```rust
 /// use rand::rngs::StdRng;
 ///
-/// use phylo::random::{RandomGenerator, RandomSource};
+/// use phylo::random::RandomGenerator;
 ///
 /// // Create a custom RNG instance
-/// let custom_rng: RandomGenerator<StdRng> = RandomGenerator::new(123);
-/// let custom_value: f64 = custom_rng.gen();
+/// let mut custom_rng: RandomGenerator<StdRng> = RandomGenerator::new(123);
```
MattesMrzik, line 27: Here, you are using `StdRng,` but below the `Pcg32`, why this discrepancy?
```rust
pub type DefaultGenerator = RandomGenerator<Pcg32>;
```

phylo/src/random/fake_random.rs
```diff
@@ -1,477 +1,305 @@
...
+            u64_values: Vec::new(),
+            u64_index: 0,
+        }
+    }
 }
 
-impl FakeGenerator {
-    /// Create a new FakeGenerator with empty value sequences,
-    /// which will default to 0, 0.0, or false if no values are provided
+unsafe impl Send for FakeRng {}
```
MattesMrzik: Is this because the compiler thinks sending FakeRng is unsafe, so we add the `unsafe` to tell the compiler that is in fact sendable?
But why would it be unsafe? its just a vector and a usize.
I also could delete this line without getting and error
I assume `Pcg32` also implements Send.

phylo/src/random/mod.rs
```diff
@@ -1,198 +1,145 @@
...
-    R: Rng + SeedableRng + Send,
-{
-    pub seed: u64,
-    pub rng: R,
-}
-
 /// A generic random number generator wrapper that can work with different RNGs.
 ///
-/// This provides a thread-safe, reusable interface for different RNG backends.
+/// This provides a reusable interface for different RNG backends (not thread-safe per se).
```
MattesMrzik: Thread safe means that we don't necessarily get the same random values for different runs, right?
So if we have a rng with seed then up to creating the new thead it should be reproducible, right? If we then move or copy the rng to the thread, shouldn't the thread reproduce the same random numbers? Perhaps if I create multiple threads and some finish faster or slower between different runs then the order of how new threads are created is different between the runs, and therefore different rngs (or different states of a rng) are passed to the threads. Am I understanding this correctly?
And this has nothing to do with the `unsafe impl Send for FakeRng {}`, right?

phylo/examples/rng_examples.rs
```diff
@@ -1,79 +1,89 @@
...
-use phylo::random::{DefaultGenerator, RandomGenerator, RandomSource};
+use phylo::random::{DefaultGenerator, RandomGenerator};
 
 fn main() {
     // 1. Using the default RNG (StdRng-based)
     let seed = 42;
-    println!("1. RNG (StdRng) with seed {seed} - reproducible:");
-    let rng = DefaultGenerator::new(seed);
+    let mut rng = DefaultGenerator::new(seed);
+    println!("1. RNG (StdRng) with seed {} - reproducible:", rng.seed());
```
MattesMrzik: Aren't we using `Pcg32`?

MattesMrzik (approved):
Looks good to me 👍 
The `RandomGenerator` wrapper is just used to access the seed?
Some of the comments are just questions asking whether I am understanding sth correctly, we can also continue this discussion in person or via teams chat.

phylo/src/random/fake_random.rs
junniest: Yes, that's a glitch, thank you for catching it! I was trying to check both `FakeRng::default()` and `FakeGenerator::default()`, but identical tests are silly. Added `FakeGenerator::default();` to the previous test instead.

phylo/src/random/mod.rs
junniest: Ah, because I changed the main code and forgot to change the examples (; Ultimately it should not matter, the example just was to show that it is possible to use different rngs with the struct. But I will change the examples to use DefaultGenerator (which is Pcg32 now).

phylo/src/random/fake_random.rs
junniest: This is curious, I think the requirement for `Send` (`R: Rng + SeedableRng + Send`) in `RandomGenerator` actually came from a Copilot suggestion that I did not check. I removed it now so the `impl` became unnecessary (at least at the moment).

However, the `Send` impl requires `unsafe` because of the trait definition, so the keyword would be necessary if the trait was required.

phylo/examples/rng_examples.rs
junniest: Yep, fixed now (:

phylo/src/random/mod.rs
junniest: The problem with making the rng accessible from multiple threads at the same time is that the order in which the threads will poll it will make a difference to what the actual outcome of the analysis will be. I think that here my terminology is not right, it's not about thread safety per se, it's rather about computational reproducibility in the context of phylogenetics. I will rephrase that.

At the moment the threads that spawn for evaluating SPR moves are deterministic -- they actually do not need/use an rng. And I think that they are created in the same order every time, as far as I understand, but this might differ between different parallel implementations, I am not sure. If we ever need the threads to use an rng, an easy fix would be to give each thread a new rng instance seeded with a value from the "main" rng which would then maintain a reproducible sequence (but for this I would need to make sure the threads are always created in the same order). Although this will break the reproducibility between parallel/sequential runs with the same seed, so I would have to make similar modifications for the sequential implementation to make sure the SPR moves are evaluated exactly the same... But for now this is not a problem because the parallelisation doesn't need an rng in the first place.

And yes, this indeed has nothing to do with that `impl` which is now gone anyways (:

junniest: Ah, again, this was an outdated comment. I added a note to the `DefaultGenerator` definition about this, but did not change the `RandomGenerator<R>` comment. I have now moved the note up, so now it says instead:
```
/// Note: This implementation is not thread-reproducible, i.e. if used in multiple threads, the sequences
/// generated may differ between runs. For phylogenetic analysis reproducibility both sequential and parallel
/// runs should produce the same results when using the same seed.
/// Users must handle this accordingly to ensure reproducibility if needed.
```

Does that make more sense to you?

phylo/src/random/fake_random.rs
junniest: Yes, of course! I will add this test, that's my bad.

And you are actually correct, this was exactly the use case (`NjTreeBuilder`) I added this for, so that I can force the sampler to pick the entries I want. The previous version was a bit more straightforward because I could just force it to pick a specific entry in the vector by index, but this is not possible when you implement the Rng traits unfortunately.

MattesMrzik (approved):

PR #128, opened 2025-12-17
Enhancement/cleanup sequences

junniest:


phylo/src/alignment/mod.rs
```diff
@@ -30,15 +30,28 @@ pub type SeqMaps = HashMap<NodeIdx, Mapping>;
...
 pub struct PairwiseAlignment {
-    pub map_x: Mapping,
-    pub map_y: Mapping,
+    pub(crate) map_x: Mapping,
+    pub(crate) map_y: Mapping,
 }
 
 impl PairwiseAlignment {
     pub fn new(map_x: Mapping, map_y: Mapping) -> PairwiseAlignment {
         debug_assert!((map_x.len() == map_y.len()) || map_y.is_empty());
```
MattesMrzik: Why do we allow the `map_y` to be empty?
Perhaps do this check always and not only in debug, or alternatively trim the longer seq to the len of the shorter.

phylo/src/alignment/sequences.rs
```diff
@@ -60,34 +140,87 @@ impl Sequences {
...
+    /// use phylo::alignment::Sequences;
+    /// use phylo::record as record;
+    ///
+    /// let records = vec![record!("seq1", None, b"A")];
+    /// let seqs = Sequences::new(records);
+    /// for record in seqs.iter() {
+    ///     println!("{}", record.id());
+    /// }
+    /// ```
+    pub fn iter(&self) -> Iter<'_> {
```
MattesMrzik: Would it also be possible to return `slice::Iter<'a, Record>` here? Then the code for `Iter`, ... would also not be needed.

MattesMrzik (approved):

phylo/src/alignment/mod.rs
junniest: Good catch. I actually noticed this but then thought it was for an implementation detail that used to exist but not anymore.

If I remember correctly I used to use the `PairwiseAlignment` struct to also store leaf maps, which would only have one map filled out, the second would be empty. At the moment, however, the leaf maps are stored just as `Mapping`s, so this check is just plain wrong. Removed it, it only checks that the maps match in length now.

To your second comment:
The check is debug because I tend to avoid assertions in production code for performance reasons, which is not the case here. This one doesn't seem to have much of a cost attached and it's part of the public API, so it should actually be a regular assert, not a debug. Plus in Rust it's acceptable for `new` to panic, so if someone tries to create an invalid `PairwiseAlignment` a panic is in order.

But it also should never be the case that something needs to be trimmed, the two maps must match in length. Trimming something will mean that we are discarding information and that could hide some nasty bugs. So panic it is!

phylo/src/alignment/sequences.rs
junniest: Good catch! I ended up implementing the iterators using CoPilot and so paid for it :D I didn't know how to implement iterators properly, but the fluff with `Iter` is clearly unnecessary, thank you.

PR #132, opened 2026-01-16
added ascii tree to illustrate spr move

MattesMrzik:


phylo/src/optimisers/spr_optimiser.rs
```diff
@@ -262,6 +262,25 @@ fn rooted_spr(tree: &Tree, prune_idx: &NodeIdx, regraft_idx: &NodeIdx) -> Result
...
+///
+/// after:                        |
+///                     -----prune_grpar-----
+///                     |                   |
+///               -------             --regraft_par--
+///               |                   |             |
+///          prune_sib                .        --prune_par--
+///                                            |           |
+///                                          prune       regraft
+/// ```
```
junniest: nitpick: Maybe we don't need the very long horizontal branch on regraft (in before) and prune_sib (in after)? Otherwise it seems like there's internal nodes that don't exist... And, maybe we should use a triangle for the undisturbed subtree: △.

Something like this:
```
/// text
/// before:                       |
///                     -----prune_grpar-----
///                     |                   |
///               --prune_par--       --regraft_par--
///               |           |       |             |
///          prune_sib      prune     △             |
///                                                 |
///                                              regraft
///
/// after:                        |
///                     -----prune_grpar-----
///                     |                   |
///                     |             --regraft_par--
///                     |             |             |
///                 prune_sib         △        --prune_par--
///                                            |           |
///                                          prune       regraft
/// 
```

junniest (approved):
Nice example!

phylo/src/optimisers/spr_optimiser.rs
MattesMrzik: The triangles are nice. I would perhaps make the branches length 2 instead of 3, then the ascii "art" matches the code as in the blen gets split in half, like so 
```
/// text
/// before:                       |
///                     -----prune_grpar-----
///                     |                   |
///               --prune_par--       --regraft_par--
///               |           |       |             |
///          prune_sib      prune     △             |
///                                              regraft
///
/// after:                        |
///                     -----prune_grpar-----
///                     |                   |
///                     |             --regraft_par--
///                prune_sib          |             |
///                                   △        --prune_par--
///                                            |           |
///                                          prune      regraft
/// 
```

junniest: yes, looks perfect!

Issue #121, opened 2025-11-11
Refactor `Tree::new()`


junniest:
The `NJTreeBuilder::build_from_distances()` accepts sequences but only needs them to create a blank tree with the correct leaf IDs. This is due to the signature of `Tree::new()`, which should be refactored. The trees should only be created by builders or by the newick parser, and the `new` method has visibility `pub(crate)`.

junniest:
Additionally, `tree.complete` should be removed, a valid tree object should always be complete, so the field does not make sense.

PR #133, opened 2026-01-19
tree almost_eq

### LINKED ISSUE #78: Implement Eq for Tree
Implement `Eq` for the `Tree` struct. You can take inspiration from:
```rust
fn compare_trees(tree: &Tree, true_tree: Tree) {
    assert_eq!(tree.root, true_tree.root);
    for node_idx in tree.preorder() {
        let current = tree.node(node_idx);
        let current_id = &current.id;
        assert_eq!(current.blen, true_tree.by_id(current_id).blen);
        if node_idx == &tree.root {
            continue;
        }
        let true_parent = true_tree.by_id(current_id);
        let parent = tree.by_id(current_id);
        assert_eq!(parent.id, true_parent.id);
    }
}
```

MattesMrzik:
Implemented a `almost_eq` method to compare trees ignoring internal representations. 
Closes #78. As we not want to directly implement `Eq` since we are still comparing `f64` fields for which `Eq` is not implemented.

### DIFF
--- a/phylo/src/tree/mod.rs
+++ b/phylo/src/tree/mod.rs
@@ -381,9 +381,9 @@ impl Tree {
     }
 
     /// Compares two trees for approximate equality, allowing for small differences in branch
-    /// lengths. Not caring about the order of nodes in the internal representation or correspondence
-    /// of [`Node`]s and [`NodeIdx`]s. The order of children of a node is ignored, i.e.,
-    /// they are compared as sets. Returning `false` in the case of `NaN` branch lengths.
+    /// lengths. This comparison does not care about the order of nodes in the internal representation
+    /// or correspondence of [`Node`]s and [`NodeIdx`]s. The order of children of a node is ignored,
+    /// i.e., they are compared as sets. It returns `false` in the case of `NaN` branch lengths.
     ///
     /// # Errors
     ///
--- a/phylo/src/tree/mod.rs
+++ b/phylo/src/tree/mod.rs
@@ -434,7 +434,7 @@ impl Tree {
                 let n2 = match n2_option {
                     Some(n) => n,
                     None => {
-                        warn!("Lineage for leaf '{}' is shorther in other tree", leaf_id);
+                        warn!("Lineage for leaf '{}' is shorter in other tree", leaf_id);
                         return false;
                     }
                 };
--- a/phylo/src/tree/mod.rs
+++ b/phylo/src/tree/mod.rs
@@ -385,7 +385,7 @@ impl Tree {
     /// or correspondence of [`Node`]s and [`NodeIdx`]s. The order of children of a node is ignored,
     /// i.e., they are compared as sets. It returns `false` in the case of `NaN` branch lengths.
     ///
-    /// # Errors
+    /// # Panics
     ///
     /// Panics if the ids of the leaves are not unique;
     pub fn almost_eq(&self, other: &Self, epsilon: f64) -> bool {
--- a/phylo/src/tree/mod.rs
+++ b/phylo/src/tree/mod.rs
@@ -167,6 +167,10 @@ impl Tree {
         partitions
     }
 
+    /// Returns true if `query` is in the subtree rooted at `node`.
+    /// This assumes that the matching between [ids](`Node::id`) and [idx](`Node::idx`) is the
+    /// same in the two trees. That means even if [`Tree::almost_eq`] returns true, this method
+    /// here can return false if you pass the two roots.
     pub(crate) fn is_subtree(&self, query: &NodeIdx, node: &NodeIdx) -> bool {
         let order = self.preorder_subroot(node);
         order.contains(query)
@@ -375,6 +379,86 @@ impl Tree {
             }
         }
         Ok(())
+    }
+
+    /// Compares two trees for approximate equality, allowing for small differences in branch
+    /// lengths. The order of children of a node is ignored, i.e., they are compared as sets.
+    /// Returning false in the case of `NaN` branch lengths.
+    ///
+    /// # Errors
+    ///
+    /// Panics if the ids of the leaves are not unique;
+    pub fn almost_eq(&self, other: &Self, epsilon: f64) -> bool {
+        if self.n != self.leaf_ids.iter().collect::<HashSet<_>>().len() {
+            panic!("Self tree has non-unique leaf ids");
+        }
+        if other.n != other.leaf_ids.iter().collect::<HashSet<_>>().len() {
+            panic!("Other tree has non-unique leaf ids");
+        }
+
+        if self.length.is_nan()
+            || other.length.is_nan()
+            || (self.length - other.length).abs() > epsilon
+        {
+            println!("Total lengths differ");
+            return false;
+        }
+        if self.n != other.n
+            || self.complete != other.complete
+            || self.nodes.len() != other.nodes.len()
+        {
+            println!("Root, n, complete, nodes.len, preorder, or postorder differ");
+            return false;
+        }
+
+        for leaf_id in &self.leaf_ids {
+            let mut n1_option = self.nodes.iter().find(|node| node.id == *leaf_id);
+            let mut n2_option = other.nodes.iter().find(|node| node.id == *leaf_id);
+
+            if n1_option.is_none() || n2_option.is_none() {
+                println!("Leaf id {} not found in both trees", leaf_id);
+                return false;
+            }
+            while let Some(n1) = n1_option {
+                let n2 = match n2_option {
+                    Some(n) => n,
+                    None => {
+                        println!("Node id {} not found in both trees", n1.id);
+                        return false;
+                    }
+                };
+                // checking the ids
+                if n1.id != n2.id {
+                    println!("Node ids differ: {} vs {}", n1.id, n2.id);
+                    return false;
+                }
+                // checking the branch lengths
+                if n1.blen.is_nan() || n2.blen.is_nan() || (n1.blen - n2.blen).abs() > epsilon {
+                    println!(
+                        "Branch lengths differ for node {}: {} vs {}",
+                        n1.id, n1.blen, n2.blen
+                    );
+                    return false;
+                }
+                // continue with the parents
+                n1_option = match n1.parent {
+                    Some(parent_idx) => Some(self.node(&parent_idx)),
+                    None => None,
+                };
+                n2_option = match n2.parent {
+                    Some(parent_idx) => Some(other.node(&parent_idx)),
+                    None => None,
+                };
+            }
+            // checking if n2_option is None as well
+            if n2_option.is_some() {
+                println!("Trees differ in structure above leaf id {}", leaf_id);
+                return false;
+            }
+        }
+
+        // not comparing Tree.dirty as it is crate internal
+        true
     }
 }
 
--- a/phylo/src/tree/mod.rs
+++ b/phylo/src/tree/mod.rs
@@ -4,6 +4,7 @@ use std::fmt::{Debug, Display};
 use anyhow::bail;
 use fixedbitset::FixedBitSet;
 use inc_stats::Percentiles;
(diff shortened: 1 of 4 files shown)


MattesMrzik:
Perhaps we want to return a `Result` instead of a `bool`? Then people can easily see how the two trees differ, by looking at the returned error msg.

MattesMrzik:
Had to change a Newick string in a test. See #134

phylo/src/tree/mod.rs
```diff
@@ -66,6 +67,7 @@ pub struct Tree {
     postorder: Vec<NodeIdx>,
     preorder: Vec<NodeIdx>,
     leaf_ids: Vec<String>,
+    // TODO: what is meant by this?
```
junniest, line 70: hahaha I agree with copilot on this, and its suggestion is correct. This was a field meant for the situation when the tree is being constructed and not nodes are all filled in, traversals are not initialised, etc. It's not a great way to handle that and I will try to refactor this out so that there's no way to get an incomplete tree in the first place. See #121.

phylo/src/tree/mod.rs
```diff
@@ -167,6 +169,7 @@ impl Tree {
         partitions
     }
 
+    /// Returns true if `query` is in the subtree rooted at `node`.
```
junniest, line 172: This could actually be a pub method, I think. In that case, it would be nice to also have a docstring example. If you don't have the time to change it I will do that in #121.

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
...
     }
+
+    /// Compares two trees for approximate equality, allowing for small differences in branch
+    /// lengths. This comparison does not care about the order of nodes in the internal representation
+    /// or correspondence of [`Node`]s and [`NodeIdx`]s. The order of children of a node is ignored,
+    /// i.e., they are compared as sets. It returns `false` in the case of `NaN` branch lengths.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the ids of the leaves are not unique;
```
junniest, line 390: I don't think that equality should check whether IDs are unique in each tree, since this is a different kind of check. It's a check on tree validity rather than on equality. I would say that you should assume that if the trees have already been created by this point (and they have), the trees themselves will be valid.

Actually, funnily enough, this made me add a new test for the newick parser and it does not fail when there are duplicate ids. I will fix that asap (#146)

phylo/src/tree/tests.rs
```diff
@@ -293,12 +293,14 @@ fn newick_parse_parentheses_around_all() {
 #[test]
 fn newick_parse_whitespace() {
     let trees = from_newick(
-        "     (     (((  (A:1   , B  :   1.0)  \n \n F:1,C:2.0   )G:1,D:3)H:+1.0  ,  E:4)   I:1)\n;\n   ",
+        "     (     (((  (A:1   , B:   1.0)  \n \n F:1,C:2.0   )G:1,D:3)H:+1.0  ,  E:4)   I:1)\n;\n   ",
```
junniest, line 296: I am a bit confused here. Why did you end up removing the whitespace? At the moment (at least for me) this test already works just fine (the parser does not include whitespace in the name).

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
...
+
+        if self.length.is_nan()
+            || other.length.is_nan()
+            || (self.length - other.length).abs() > epsilon
+        {
+            warn!("Tree lengths differ: {} vs {}", self.length, other.length);
+            return false;
+        }
+        if self.n != other.n
+            || self.complete != other.complete
```
junniest, line 407: I will remove this flag, so you can remove this check. A created tree should always be complete and valid.

The other thing is that `self.n` must be equal to `self.nodes.len()`, so this will be a duplicate check which can also be removed.

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
...
+    pub fn almost_eq(&self, other: &Self, epsilon: f64) -> bool {
+        if self.n != self.leaf_ids.iter().collect::<HashSet<_>>().len() {
+            panic!("Self tree has non-unique leaf ids");
+        }
+        if other.n != other.leaf_ids.iter().collect::<HashSet<_>>().len() {
+            panic!("Other tree has non-unique leaf ids");
+        }
+
+        if self.length.is_nan()
+            || other.length.is_nan()
```
junniest, line 400: This check is unnecessary because of the way that Rust handles NaNs. Any combination of (f64 - NaN), (NaN - f64), (NaN - NaN) results in a NaN, any further action (e.g. `abs()`) results in a NaN and any comparisons against NaN will be always false (even NaN == NaN is false). So you can leave the last line of this comparison as the only one.

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
...
+                other.n,
+                self.complete,
+                other.complete,
+                self.nodes.len(),
+                other.nodes.len()
+            );
+            return false;
+        }
+
+        // For each leaf, traverse up to the root and compare the nodes along the way
```
junniest, line 422: In the spirit of making such checks/comparisons as efficient as possible, I think it makes sense to do the most basic/fast checks first before going through the leaves.

So to me, it would make sense to first do these two:
a) compare the leaf sets and return false if they differ (which is a relatively quick check);
b) get the RF distance and return false is it is > 0 (which is slower, and specifically doesn't check that the leaf ids match exactly, one tree can be a subset of another).

Then, you could go through the leaf nodes of the one of the trees, which also means that you are certain that they have the exact same leaves.

My other concern here is that you will be walking the whole length of the tree for each leaf, meaning that you will check the internal nodes more than once for sure. I would recommend rather switching to something like postorder traversal instead.

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
...
+            while let Some(n1) = n1_option {
+                let n2 = match n2_option {
+                    Some(n) => n,
+                    None => {
+                        warn!("Lineage for leaf '{}' is shorter in other tree", leaf_id);
+                        return false;
+                    }
+                };
+                // checking the ids
+                if n1.id != n2.id {
```
junniest, line 442: Ok, this is tricky. Currently if we parse newick trees and the provided tree doesn't have some or all internal node ids, they will not be added. In the context of ASR, this wouldn't make much sense, since then the id will be black in output and there will be no way to connect the MASA to the internal nodes.

I guess this is not a problem for when the MASA is provided as input since there's no way to reconstruct the tree from it anyway, the tree has to be provided. But in the case when we don't initialise with a MASA but only with an MSA and build/parse a tree, we likely need to fill out the internal node ids (even if it's a representation of NodeIdx), to make sure that the output can be matched properly. I added an issue for this (#147).

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
...
+                    warn!(
+                        "Node ids differ: '{}' vs '{}', in the lineage for leaf '{}'",
+                        n1.id, n2.id, leaf_id
+                    );
+                    return false;
+                }
+                // checking the branch lengths
+                if n1.blen.is_nan() || n2.blen.is_nan() || (n1.blen - n2.blen).abs() > epsilon {
```
junniest, line 450: Here again, any NaNs with always fail the last comparison, so you only need that one for this check to work.

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
...
+                    None => None,
+                };
+            }
+            if n2_option.is_some() {
+                warn!("Lineage for leaf id '{}' is longer in other tree", leaf_id);
+                return false;
+            }
+        }
+
+        // not comparing Tree.dirty as it is crate internal
```
junniest, line 473: Uuuugh, I think the `dirty` bitset should not be a part of the tree, now that I think about it... But I am not sure what would be a better place for it.

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
...
+                n1_option = match n1.parent {
+                    Some(parent_idx) => Some(self.node(&parent_idx)),
+                    None => None,
+                };
+                n2_option = match n2.parent {
+                    Some(parent_idx) => Some(other.node(&parent_idx)),
+                    None => None,
+                };
+            }
+            if n2_option.is_some() {
```
junniest, line 467: And the trees need to be rooted the exact same way, which can be both good and bad. I am not sure if it's worth trying to compare trees with different rootings, but if we expect the rooting to be the same I think it should be said in the doctring.

phylo/src/tree/mod.rs
```diff
@@ -376,6 +379,100 @@ impl Tree {
         }
         Ok(())
     }
+
+    /// Compares two trees for approximate equality, allowing for small differences in branch
+    /// lengths. This comparison does not care about the order of nodes in the internal representation
+    /// or correspondence of [`Node`]s and [`NodeIdx`]s. The order of children of a node is ignored,
+    /// i.e., they are compared as sets. It returns `false` in the case of `NaN` branch lengths.
```
junniest, line 386: The last sentence isn't necessary, imo (see my comments about NaNs below). NaNs should not appear in any case, but if they do the whole structure is invalid anyway.

phylo/src/tree/tests.rs
```diff
@@ -739,3 +741,94 @@ fn rf_distance_against_raxml() {
...
+#[test]
+fn almost_eq_trees_with_wrong_leaf_id() {
+    let tree1 = tree!("((A:1,B:2)E:5,(C:3,D:4)F:6)G:7;");
+    let tree2 = tree!("((D:4.1,C:2.9)F:6.1,(B:2.1,T:0.9)E:4.9)G:7;");
+    let epsilon = 100.0;
+    assert!(!tree1.almost_eq(&tree2, epsilon));
+}
+
+#[test]
+fn almost_eq_trees_with_wrong_internal_id() {
```
junniest, line 829: This is curious. Technically, if we have the same structure (same leaf clustering) the internal node names shouldn't matter. I am not sure how to make that into a rule here and whether we should bother with it at all, but it might pop up eventually.

junniest (commented):
This is a much more complex problem (comparing trees) than it looked on the surface :D I think we still need to think how to tackle it consistently and properly.

phylo/src/tree/tests.rs
MattesMrzik: Thats weird. With this PR the ids are compared which are "B  " for the first tree and "B" for the second (before this PR ids where not checked in the test). See #136 that fixed that. So when #136 is merged this whitespace can be added back.

phylo/src/tree/mod.rs
MattesMrzik: No its not. `self.n` is the number of leaves not nodes. If we assume that the tree struct is valid, is rooted and binary, then if the `n`s are equal it follows that the nodes.length also match.

phylo/src/tree/mod.rs
MattesMrzik: Thanks:)

phylo/src/tree/mod.rs
MattesMrzik: I think i did it on a per lineage basis and not postorder since i didnt want to assume that the matching between node ids and node idxs are consistent between the two trees. And i wasnt sure how to compare the trees without relying on that assumption, so i implemented this inefficient (but working) algo.

MattesMrzik:
- Perhaps we wait until we get `rerooting`, then we can reroot to the same edge of the two compared trees and then do the comparison. 
-  Currently, we are using the ids (type String) to connect tree nodes with sequences. If we have a phylo info with a masa (ie internal seqs), then it makes sense to me that we include ids in the comparison.

PR #135, opened 2026-01-20
Msa file io and record sort

MattesMrzik:
Sorting records in MSA and MASA by id, then desc, then seq when outputting with fmt or to a file. 
Implemented write_msa_to_file.

### DIFF
--- a/phylo/src/alignment/mod.rs
+++ b/phylo/src/alignment/mod.rs
@@ -152,6 +152,8 @@ impl Display for MSA {
             let aligned_seq = aligned_seq!(seq_map, record.seq());
             aligned_records.push(record!(id, record.desc(), &aligned_seq));
         }
+        // sort records by id to have a consistent output
+        aligned_records.sort_by(|a, b| a.id().cmp(b.id()));
         write!(f, "{}", Sequences::new(aligned_records))
     }
 }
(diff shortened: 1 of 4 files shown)


MattesMrzik:
Would it perhaps be nicer to write a general method that takes in any type T that implements `Display` and a path and write a file? Then the very similar methods (for MSA/MASA at least) could be collapsed.

MattesMrzik:
Perhaps sort first by leaves, internals then root like @junniest mentioned: 
> One question here is whether we would like to keep the original order of sequences (or at least have the leaves printed first, then internal nodes, then root?).I am not sure what is best here. Trying to maintain the order which we read the sequences in seems like to much work. But potentially first printing the leaves, then the internal nodes and the root last could be reasonable?

MattesMrzik:
Perhaps sorting like this: 
```rust
let both_maps = both_maps.sorted_by_key(|(node_idx, _)| {
    &self.idx_to_id[usize::from(**node_idx)]
});
```
is nicer.

MattesMrzik:
Two questions: 

1. Do we want to output Gaps that are due to an insertion in a different subtree as a different character? Like in IndelMap? I think `*` was used for that. 
2. Do we want to output characters that were inserted as lowercase?

PR #138, opened 2026-01-20
Add debug print for test results in workflow

MattesMrzik:
Pulled the cmd 
`cargo test --manifest-path ./phylo/Cargo.toml -- -Z unstable-options --format json --report-time | cargo2junit > results.xml`
apart: 
```
cargo test --manifest-path ./phylo/Cargo.toml -- -Z unstable-options --format json --report-time > cargo_test.out
cat cargo_test.out
cargo2junit < cargo_test.out > results.xml
cat results.xml
```
Therefore, even if publishing the results fails we can look at std::out and see which tests failed.

PR #139, opened 2026-01-21
Fix failing tests not shown

MattesMrzik:
This PR is just for testing purposes.


MattesMrzik:
Will close this PR since it was only for testing the workflow, see #138 for the actual PR.

PR #140, opened 2026-01-23
Simulate ancestral alignments under the TKF and substitutions models

MattesMrzik:
This PR implements an `AlignmentSimulation` trait which simulates an ancestral alignment and implements it for the TKF91 and TKF92 models and substitution models. It also introduces a utility for removing non surviving homology paths and ancestral alignment transformation into standard leaf-only alignments.
Key Changes
- TKF Indel MSA Simulation
- Full Evolutionary Simulation: Added TKFMSASimulator which combines the indel process with a substitution model to produce complete, character-based alignments.
- Substitution Simulator: Introduced a standalone SubstitutionSimulator for generating alignments with substitutions only.
- Ancestral Alignment Enhancements:
    - Added remove_extinct_columns to clean up alignments where ancestral characters do not survive to any leaves.
    - Implemented into_alignment to transform ancestral alignments (containing internal nodes) into standard alignments.
- Test: Added tests covering mutation constraints, Dollo’s constraint compliance, and fragment capping logic.


MattesMrzik:
Some questions:
- `prune_empty_leaves`: perhpas we want this method action on `PhyloInfo`, or as a `fn` in the trait `AlignmentSimulation`?
- In the `Simulators`, perhaps we want the `RNG`s mutability not hidden behind a `RefCell` and instead change the `fn`s that use it take a mutable reference to self.

phylo/src/tkf_model/simulate_msa/sim_tkf_indel_msa.rs
```diff
@@ -0,0 +1,931 @@
...
+            }
+        }
+        let immortal_prob = 1.0 - cumulative_prob;
+        *self.cumulative_logl.borrow_mut() += immortal_prob.ln();
+        self.max_insertion_length
+    }
+
+    /// After [simulation of the links](`Self::build_msa_links`) is complete, this function
+    /// calls [`Self::append_link_to_msa`] for every root link to build the final MSA and fragmentation.
+    fn links_to_msa<AA: AncestralAlignment>(&self, links: &Vec<TKFLink>) -> (AA, Fragmentation) {
```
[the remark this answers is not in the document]
MattesMrzik: impled it

phylo/src/tkf_model/simulate_msa/sim_tkf_msa.rs
```diff
@@ -0,0 +1,379 @@
...
+                )
+            });
+        // calling 'from_aligned_with_ancestral' removes all-gap cols. There should be none,
+        // so asserting here that the msa did indeed not shrink.
+        assert_eq!(
+            masa.len(),
+            aln_len,
+            "Final MSA length should match the indel MSA, but it does not. \
+             Please report this at {REPORT_ISSUES_URL}."
+        );
```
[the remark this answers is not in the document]
MattesMrzik: This is incorrect. Extinct columns mean gaps at leaves, there still are chars at the internal nodes.

phylo/src/alignment/mod.rs
```diff
@@ -124,11 +125,80 @@ pub trait AncestralAlignment: Alignment {
...
+            .enumerate()
+            .filter_map(|(col_idx, survives)| (!survives).then_some(col_idx))
+            .collect();
+        if !extinct_cols.is_empty() {
+            warn!(
+                "Columns {} go extinct in all leaf sequences. \
+                Consider calling `remove_extinct_columns` on the alignment.",
+                extinct_cols.iter().format(", ")
+            );
+        }
```
junniest: Hmmm. Should this maybe be refactored to a separate method? It bloats this method quite a bit and makes it harder to read.

Another thought is whether this check is necessary? Since we call `Sequences::remove_gap_cols` before the ancestral alignment is constructed, there shouldn't be any MASA cols that are extinct, which means there shouldn't be any once we create the ancestral alignment out of the input.

Are extinct columns still possible here after that call? Do we have a test that can show that?

Since I mention this method, should we rename the `Sequences::remove_gap_cols` to something along the lines of `remove_extinct_columns` as well? That seems to be a more meaningful name.

phylo/src/alignment/mod.rs
```diff
@@ -124,11 +125,80 @@ pub trait AncestralAlignment: Alignment {
...
+        Ok(ancestral_alignment)
     }
     /// Constructs an ancestral alignment instance from aligned sequences and a phylogenetic tree. Is called
     /// by the default implementation of [`Self::from_aligned_with_ancestral`].
     fn from_aligned_with_ancestral_unchecked(all_seqs: Sequences, tree: &Tree) -> Self;
+    /// Removes columns from the alignment where the ancestral character goes extinct, i.e. columns
+    /// where all leaf sequences have a gap and returns a boolean mask indicating which columns were
+    /// removed, ie `true` for columns that were kept and `false` for columns that were removed.
+    fn remove_extinct_columns(&mut self) -> Vec<bool>;
+    /// Transforms self into a regular alignment, i.e., without ancestral sequences.
```
junniest: I think this should rather maybe be an implementation of the `From` trait on the actual MASA struct, just to be on the safe side?..

I have to say that this implementation doesn't sit right with me. It takes a properly defined alignment, deconstructs it, recreates the mappings and creates an MSA. It feels superfluous and a potential source of error. 

There also has to be a good reason to want to transform `AncestralAlignment` into an `Alignment`, and I am not sure we should be giving a blanket implementation. If we move this into the `MASA` struct, we will control how it works better. Or, rather than implementing `From`, we could have a constructor for the MSA, something like `MSA::with_MASA()`, which is the idiomatic way to call constructors. Then it would also be obvious that we are creating a new, independent object.

Finally, if this is the only way it is possible to transform an MASA into an MSA (going through the Sequence object, recreating aligned sequences and calling `from_aligned_unchecked`), I think something is seriously wrong with our API and we should discuss how to avoid all the unnecessary transformations.

phylo/src/alignment/mod.rs
```diff
@@ -578,6 +648,67 @@ impl AncestralAlignment for MASA {
             internal_alignments: HashMap::<NodeIdx, PairwiseAlignment>::new(),
         }
     }
+
+    fn remove_extinct_columns(&mut self) -> Vec<bool> {
```
junniest: This, at the very least, needs a docstring that explains what is being done and what is being returned.

I would probably also make this one `(mut self) -> Self` or something like that, to make it clear that this will rewrite the alignment. It doesn't cost us anything, but I think it would make the point of the method more clear.

The result type also felt confusing to me to be honest. Is it really necessary for anything other than logging and warning the user?

phylo/src/alignment/mod.rs
```diff
@@ -578,6 +648,67 @@ impl AncestralAlignment for MASA {
...
+        // Determine columns where the ancestral character did not go extinct
+        let keep_cols = surviving_columns_mask(self);
+        // If no character goes extinct, we don't have to update the alignment
+        if keep_cols.iter().all(|b| *b) {
+            return keep_cols;
+        }
+        if keep_cols.iter().all(|b| !*b) {
```
junniest: I would definitely replace the `Vec<bool>` with `fixedbitset::FixedBitSet` or something of the sort. The you could call `FixedBitSet::is_full()` to check if all the columns have to be gone. I would prefer that over iterating through a vector of booleans, especially since its a lot easier to read and understand what it is doing.

phylo/src/alignment/tests.rs
```diff
@@ -574,3 +574,77 @@ fn update_ancestral_map_leaf() {
...
+    ];
+
+    let seqs = Sequences::new(records).unwrap();
+    let mut masa = MASA::from_aligned_with_ancestral(seqs, &tree).unwrap();
+
+    // sanity checks before removal (the second column is removed when the MASA is created)
+    assert_eq!(masa.len(), 6, "lens differ for initial length");
+    // after removing the two extinct columns the length should be 4
+    masa.remove_extinct_columns();
+    assert_eq!(masa.len(), 4, "lens differ after removal");
```
junniest: I think I am not getting something here... So in this case the provided ancestral alignment has a column where a character is inferred as an ancestor but then dies out but remains in the data? I guess that could happen in a simulation or if a different ASR algo is bugged? Since it doesn't really make much sense to keep that column...

phylo/src/alignment/tests.rs
```diff
@@ -574,3 +574,77 @@ fn update_ancestral_map_leaf() {
...
+        record_wo_desc!("B", b"C--CT--"),
+        record_wo_desc!("C", b"G--CT-G"),
+        record_wo_desc!("I", b"T--CTAG"),
+        record_wo_desc!("R", b"A-AATGG"),
+    ];
+
+    let seqs = Sequences::new(records).unwrap();
+    let mut masa = MASA::from_aligned_with_ancestral(seqs, &tree).unwrap();
+
+    // sanity checks before removal (the second column is removed when the MASA is created)
```
junniest: Should we also check that the original sequences were a different length?

phylo/src/alignment/mod.rs
```diff
@@ -610,12 +741,38 @@ impl MASA {
...
+    /// Simulates an ancestral alignment, i.e. an alignment that also contains sequences for
+    /// the ancestral nodes of the tree. By pure chance, some columns may go extinct, i.e.,
+    /// the homology path in that column reaches no leaf, consider calling
+    /// [`AncestralAlignment::remove_extinct_columns`] on the resulting alignment.
+    fn simulate_ancestral_alignment<AA: AncestralAlignment>(&self) -> AA;
+    /// Simulates an alignment, i.e., returns only the leaf sequences (without ancestral sequences).
+    /// It default implementation calls [`Self::simulate_ancestral_alignment`] and then transforms
+    /// the resulting ancestral alignment into a regular alignment using
+    /// [`AncestralAlignment::into_alignment`].
+    fn simulate_alignment<A: Alignment>(&self) -> A {
```
junniest: This really makes me wonder if it shouldn't be returning `PhyloInfo` rather than an alignment/ancestral alignment, since neither of those really make sense in isolation (i.e. without the tree).

phylo/src/alignment/mod.rs
```diff
@@ -610,12 +741,38 @@ impl MASA {
...
+        self.simulate_ancestral_alignment::<MASA>()
```
junniest: This looks strange to me. It requires a trait but also the MASA struct? So we are still bound to our struts despite having the traits.

phylo/src/substitution_models/simulate_msa.rs
```diff
@@ -0,0 +1,225 @@
...
+    ///   [`set_missing_tree_node_ids`](`crate::phylo_info::set_missing_tree_node_ids`)).
+    pub fn new<Q: QMatrix>(
+        model: SubstModel<Q>,
+        tree: Tree,
+        rng: RandomGenerator<R>,
+        alignment_length: usize,
+    ) -> Result<Self> {
+        let tree = set_missing_tree_node_ids(&tree)?;
+        let root_dist = WeightedIndex::new(model.qmatrix.freqs().as_slice()).unwrap_or_else(|e| {
+            panic!(
```
junniest: Out of curiosity, can this ever happen? Unless somehow `DVector<f64>` is bugged...

phylo/src/substitution_models/simulate_msa.rs
```diff
@@ -0,0 +1,225 @@
...
+            let mut column_dists = Vec::with_capacity(p.ncols());
+            for col in 0..p.ncols() {
+                let column = p.column(col);
+                let weighted_index = WeightedIndex::new(column.as_slice()).unwrap_or_else(|e| {
+                    panic!(
+                        "Getting WeightedIndex from probability transition matrix failed. This should never happen. \
+                        Please report this at {REPORT_ISSUES_URL}. \
+                        Error: {e}"
+                    )
+                });
```
junniest: Since this is repeated at least twice, I suggest it should be a separate function.

phylo/src/substitution_models/simulate_msa.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub fn alignment_length(&mut self, length: usize) {
+        self.alignment_length = length;
+    }
+
+    pub(crate) fn simulate_ancestral_alignment_with_length<AA: AncestralAlignment>(
+        &self,
+        alignment_length: usize,
+    ) -> AA {
+        if alignment_length == 0 {
+            warn!("Alignment length is set to 0, will produce an empty alignment");
```
junniest: This should cause an immediate return, imo.

phylo/src/substitution_models/simulate_msa.rs
```diff
@@ -0,0 +1,225 @@
...
+        })
+    }
+}
+
+impl<R> AlignmentSimulation for SubstitutionSimulator<R>
+where
+    R: Rng + SeedableRng + RngCore,
+{
+    fn simulate_ancestral_alignment<AA: AncestralAlignment>(&self) -> AA {
+        self.simulate_ancestral_alignment_with_length(self.alignment_length)
```
junniest: is the argument necessary if we're passing self anyways?

phylo/src/substitution_models/simulate_msa.rs
```diff
@@ -0,0 +1,225 @@
...
+    use crate::tree;
+
+    use super::*;
+
+    #[test]
+    fn test_substitution_simulator() {
+        // GTR with chosen freqs and rate parameters
+        let model = SubstModel::<GTR>::new(&[0.3, 0.2, 0.2, 0.3], &[0.8, 1.2, 0.9, 1.1, 0.7]);
+        let tree = tree!("((A:2.0,B:2.0)AB:2.0,(C:2.0,D:2.0)CD:2.0)R;");
+        let rng = DefaultGenerator::new(123);
```
junniest: Unless we're testing reproducibility, I think we might want to use `FakeGenerator`?

phylo/src/substitution_models/simulate_msa.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /// Sets the alignment length for the simulation.
+    ///
+    /// A zero `alignment_length` is accepted, but simulation will emit a warning and return an empty alignment.
+    pub fn alignment_length(&mut self, length: usize) {
+        self.alignment_length = length;
+    }
+
+    pub(crate) fn simulate_ancestral_alignment_with_length<AA: AncestralAlignment>(
```
junniest: nitpick: I think this method should be split up into smaller ones that deal with separate concerns.

phylo/src/substitution_models/simulate_msa.rs
```diff
@@ -0,0 +1,225 @@
...
+        &self,
+        alignment_length: usize,
+    ) -> AA {
+        if alignment_length == 0 {
+            warn!("Alignment length is set to 0, will produce an empty alignment");
+        }
+        let mut sequences: HashMap<NodeIdx, Vec<usize>> = HashMap::with_capacity(self.tree.len());
+
+        let mut rng = self.rng.borrow_mut();
```
junniest: Should the rng be a RefCell here? I guess potentially the rng could be used to run several simulations simultaneously, but that will, for sure, mess with the reproducibility, which I think we might want to avoid.

phylo/src/substitution_models/simulate_msa.rs
```diff
@@ -0,0 +1,225 @@
...
+        );
+        // no gaps in this simulation, so all sequences should have the same length as the alignment
+        for seq in alignment.seqs() {
+            assert_eq!(seq.seq().len(), alignment.len());
+        }
+    }
+
+    #[test]
+    fn test_reproducibility() {
+        // GTR with same parameters to ensure reproducibility across RNGs
```
junniest: Technically, we check the rng reproducibility in the rng tests, so this, potentially, isn't necessary. Although I think that it's a good test to have, but not as a unit test but rather as an integration test, what do you think?

phylo/src/tkf_model/simulate_msa/fragmentation.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// Returns the number of fragments.
+    pub fn len(&self) -> usize {
+        self.right_exclusive_boundaries.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.right_exclusive_boundaries.is_empty()
+    }
+
+    pub(super) fn fragmentation_works_with_ancestral_alignment<AA: AncestralAlignment>(
```
junniest: I am not sure what this does. Is it for situations when a fragmentation is provided and needs to be checked? Could it be fixed if the fragmentation doesn't match or is it a critical failure and can it happen "naturally"?

phylo/src/tkf_model/simulate_msa/sim_result.rs
```diff
@@ -0,0 +1,392 @@
...
+
+impl<AA: AncestralAlignment> TKFSimulationResult<AA> {
+    pub fn remove_extinct_columns(&mut self) -> Result<()> {
+        let keep_col_mask = self.masa.remove_extinct_columns();
+        self.fragmentation.remove_cols(&keep_col_mask)?;
+        self.fragmentation
+            .fragmentation_works_with_ancestral_alignment(&self.masa)
+    }
+
+    pub fn prune_empty_leaves(&mut self) -> Result<()> {
```
junniest: Oh damn. So the simulation would also adjust the tree to match? I am not sure if this is wise, and in any case then the simulation should return `PhyloInfo`, so that the MSA/MASA is strictly coupled to the tree. Otherwise one might try to use the tree they passed to the simulator assuming that it matches the final result and *sometimes* get an error when something like this happens (i.e. the new tree is different from the one used in the simulation).

A different question is whether we should simulate a fixed number of columns and then throw some out after checks or simulate until we get the required number of columns instead, which would remove the need for changing the tree during simulation.

phylo/src/tkf_model/tkf91.rs
```diff
@@ -92,6 +101,20 @@ impl Display for TKF91IndelModel {
     }
 }
 
+impl FragmentSampler for TKF91IndelModel {
+    /// In TKF91 every residue is its own independent link, so the fragment
+    /// length is always 1 and its log-probability is ln(1) = 0.0.
+    fn sample_fragment_length<R: rand::Rng>(&self, _rng: &mut R) -> (usize, f64) {
```
junniest: Shouldn't we use our in-house `RandomGenerator<R>` here? Since this is what likely be passed from the rest of the code. I know it's not used here, but its likely relevant for the TKF92 simulation.

phylo/src/tkf_model/tkf92.rs
```diff
@@ -128,6 +142,27 @@ impl Display for TKF92IndelModel {
     }
 }
 
+impl FragmentSampler for TKF92IndelModel {
+    /// Samples fragment length from a geometric distribution parameterised by `r`.
+    /// `r` is the probability that the next residue continues the current fragment,
+    /// so `1 - r` is the stopping (success) probability of the geometric draw.
+    /// Returns `(length, log_probability)`.
+    fn sample_fragment_length<R: rand::Rng>(&self, rng: &mut R) -> (usize, f64) {
```
junniest: As per the TKF91 comment, I think we should use `crate::random::RandomGenerator` here, since that will be what the rest of the code is using. This would allow using `FakeGenerator` for testing.

phylo/src/tkf_model/simulate_msa/sim_tkf_indel_msa.rs
```diff
@@ -0,0 +1,954 @@
...
+/// let max_insertion_length = 50;
+/// let simulator = TKFIndelMSASimulator::new(
+///     TKF92IndelModel::new(lambda, mu, r),
+///     tree,
+///     DefaultGenerator::new(seed),
+///     max_insertion_length,
+/// )
+/// .unwrap();
+/// let result = simulator.simulate_with_fragments::<MASA>();
+/// assert_eq!(result.masa.seq_count(), 3);
```
junniest: But since this is using the proper random generator, wouldn't it be possible for one of the leaves to go extinct which would make this assertion fail? I understand that it's extremely unlikely, but it still worries me.

phylo/src/tkf_model/simulate_msa/sim_tkf_indel_msa.rs
```diff
@@ -0,0 +1,954 @@
...
+    fn sample_fragment_length<R: Rng>(&self, rng: &mut R) -> (usize, f64);
+}
+
+/// Computes the expected length of the root sequence (number of characters)
+/// under the TKF model if the process has reached stationarity.
+pub trait ExpectedRootLength {
+    fn expected_root_length(&self) -> f64;
+}
+
+/// Since these sequences are built incrementally, we can't use the [`Sequences`] which hold immutable [records](`record`).
```
junniest: This should probably be a warranted change in the API for `Sequences` so that we don't use multiple different structures to represent the same thing. I guess the issue is that we cannot modify the sequence of `bio::Record`?

I guess in this scenario the description field is irrelevant, but what worries me is that we can start losing sequence descriptions during some of the transformations.

phylo/src/tkf_model/simulate_msa/sim_tkf_indel_msa.rs
```diff
@@ -0,0 +1,954 @@
...
+
+    /// Simulates the indel process and returns the alignment, fragmentation, and log-likelihood of the simulated MSA
+    /// under the model, which is accumulated during the simulation.
+    fn simulate_with_fragments_and_logl<AA: AncestralAlignment>(
+        &self,
+    ) -> (TKFSimulationResult<AA>, f64) {
+        *self.cumulative_logl.borrow_mut() = 0.0;
+        let links = self.build_msa_links();
+        let (masa, fragmentation) = self.links_to_msa(&links);
+        let logl = *self.cumulative_logl.borrow();
```
junniest: nitpick: it frustrates me that this assignment comes before result but after it in the return statement (:

phylo/src/tkf_model/simulate_msa/sim_tkf_indel_msa.rs
```diff
@@ -0,0 +1,954 @@
...
+        let mut non_homolog_sum = 0.0;
+        for n in 1..100 {
+            non_homolog_sum += non_homolog_prob(n, lambda, mu, beta, time);
+        }
+
+        assert_relative_eq!(non_homolog_sum, non_homolog_integrated, epsilon = 1e-10);
+    }
+
+    #[test]
+    fn tkf92_indel_simulate() {
```
junniest: I like the idea of testing multiple seeds, but this really an integration test then, not a unit test. I think we should be using `FakeGenerator` for simple unit tests with coverage and do something like this as a larger integration test.

phylo/src/tkf_model/simulate_msa/sim_tkf_msa.rs
```diff
@@ -0,0 +1,405 @@
...
+        let dummy_len = 0;
+        let subst_sim = SubstitutionSimulator::new(subst_model, tree, rng, dummy_len)?;
+        Ok(Self {
+            indel_sim,
+            subst_sim,
+        })
+    }
+
+    /// Sets the constraint for the [`RootLength`] used for the simulation.
+    pub fn root_length(&mut self, root_length: RootLength) -> &mut Self {
```
junniest: hm, why are these all references? wouldn't it make sense to just pass the object itself?

phylo/src/tkf_model/simulate_msa/sim_result.rs
```diff
@@ -0,0 +1,392 @@
...
+                }
+                tree.root = child;
+            }
+            _ => break,
+        }
+    }
+}
+
+/// Rebuilds a clean tree from a (possibly mutated) tree, dropping any nodes that are no
+/// longer connected to the root. Serialization to Newick and re-parsing re-computes all
```
junniest: Uhhhh, I understand why this is the easiest way to handle tree creation in this scenario, but this seems wrong somehow... using serialisation as a tool for anything other than output.

Would there be an easy way to create the tree without resorting to newick that we could add?

phylo/src/tkf_model/simulate_msa/sim_result.rs
```diff
@@ -0,0 +1,392 @@
...
+                    Some(*node_idx)
+                }
+                _ => None,
+            })
+            .collect()
+    }
+
+    /// Returns all sequences (leaf and ancestral) in aligned form, mirroring
+    /// [`Display`](`std::fmt::Display`) for MASA. Node ids are resolved through the tree,
+    /// since `idx_to_id` is private to the alignment module.
```
junniest: This also somehow feels wrong that you have to resolve node ids manually. Could we implement a helper that isn't relying on private indices, or work around this a different way?

MattesMrzik:
(An alternative would be to have this code as a separate crate that uses the lib via its API)

Issue #123, opened 2025-11-13
TKF92: update README and examples


MattesMrzik:
If #115 is implemented update the crate README and the examples in the example dir

PR #157, opened 2026-02-24
update readme after tkf merge

MattesMrzik:
- Addresses the README part of #123 
- bumps MSRV to 1.88 since the crate time requires it after updating the rust tool chain
- update some minor outdated stuff

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,9 @@ A high-performance Rust library for phylogenetic analysis and multiple sequence 
 
 ## Current Functionality
 
-- **Maximum Likelihood Phylogenetic Analysis**: Efficient implementation of phylogenetic tree inference using SPR moves using likelihood or parsimony cost functions;
+- **Maximum Likelihood Phylogenetic Analysis**: Efficient implementation of phylogenetic tree inference using SPR or NNI moves using likelihood or parsimony cost functions;
 - **Multiple Sequence Alignment (MSA)**: Support for Multiple Sequence Alignment using the IndelMaP algorithm ([paper]( https://academic.oup.com/mbe/article/41/7/msae109/7688856 ), [python implementation]( https://github.com/acg-team/indelMaP ));
-- **Sequence Evolution Models**: Support for various DNA (JC69, K80, TN93, HKY, GTR) and protein (WAG, HIVB, BLOSUM62) substitution models as well as the Poisson Indel Process (PIP) ([paper]( https://www.pnas.org/doi/10.1073/pnas.1220450110 )) model;
+- **Sequence Evolution Models**: Support for various DNA (JC69, K80, TN93, HKY, GTR) and protein (WAG, HIVB, BLOSUM62) substitution models as well as the Poisson Indel Process (PIP) ([paper]( https://www.pnas.org/doi/10.1073/pnas.1220450110 )) model and the [TKF91](https://link.springer.com/article/10.1007/BF02193625)/[TKF92](https://link.springer.com/article/10.1007/BF00163848) models;
 - **High Performance**: Optimised tree search with optional parallel processing capabilities.
 
 ## Getting Started
@@ -31,7 +31,7 @@ Once published on crates.io, you'll be able to use:
 phylo = "0.1.0"
 ```
 
-**Minimum Supported Rust Version**: 1.82.0
+**Minimum Supported Rust Version**: 1.88.0
 
 MSRV detected using [`cargo-msrv`]( https://github.com/foresterre/cargo-msrv ).
 
@@ -72,6 +72,11 @@ This crate supports several optional features:
 - `par-regraft-manual`: Enable manual parallel regrafting control;
 - `precomputed-test-results`: Speed up test runs with precomputed results (for local development).
 
+Test related features:
+
+- `recompute-brute-force-asr`: Force re-computation of brute-force ancestral sequence re-estimation in tests instead of loading precomputed results
+- `multithread-brute-force-asr`: Parallelise brute-force ancestral sequence re-estimation computation in tests
+
 Enable features in your Cargo.toml:
 
 ```toml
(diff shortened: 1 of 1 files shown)


MattesMrzik:
Some questions: 
- perhaps just dont mention the test brute force features?
- The `High Performance` mentions tree search again after it has been mentioned by `Maximum Likelihood Phylogenetic Analysis`, this feels kinda duplicate to me. But perhaps its fine if we want to put an emphasis on the `high performance`. What do you mean by `Optimised tree search`?

junniest:
> * perhaps just dont mention the test brute force features?

I think we should mention them, we should keep the docs as complete as possible. I will likely remove the damn `precomputed-test-results` feature once I move the tests out of the unit test category, but I think these two will stay. It doesn't really make sense to hide them, although I wish there was a way to add test-only features.

> * The `High Performance` mentions tree search again after it has been mentioned by `Maximum Likelihood Phylogenetic Analysis`, this feels kinda duplicate to me. But perhaps its fine if we want to put an emphasis on the `high performance`. What do you mean by `Optimised tree search`?

Yeah, that text was my very hasty attempt at creating an outline. It was meant to indicate that we do not do things stupidly (i.e. include intermediate cache to speed things up) and also include parallelisation :D I am not sure how to phrase this better.

README.md
```diff
@@ -10,9 +10,9 @@ A high-performance Rust library for phylogenetic analysis and multiple sequence
 
 ## Current Functionality
 
-- **Maximum Likelihood Phylogenetic Analysis**: Efficient implementation of phylogenetic tree inference using SPR moves using likelihood or parsimony cost functions;
+- **Maximum Likelihood Phylogenetic Analysis**: Efficient implementation of phylogenetic tree inference using SPR or NNI moves using likelihood or parsimony cost functions;
```
junniest, line 13: I think we should make it clear that the TKF models only work with NNI search, not with SPRs. Not sure where the best place would be for this.

README.md
```diff
@@ -31,7 +31,7 @@ Once published on crates.io, you'll be able to use:
 phylo = "0.1.0"
 ```
 
-**Minimum Supported Rust Version**: 1.82.0
+**Minimum Supported Rust Version**: 1.88.0
 
 MSRV detected using [`cargo-msrv`]( https://github.com/foresterre/cargo-msrv ).
```
junniest, line 36: See my messages in Teams, but we might want to remove this line, since the version we'll go to isn't what cargo-msrv suggests.

README.md
```diff
@@ -112,11 +116,13 @@ This is a new library that is currently in active development. Contributions are
 
 Please read our [contributor guide]( CONTRIBUTING.md )!
 
-### Current Contributors:
+### Contributors:
```
junniest, line 119: Nice restructuring, good call.

README.md
```diff
@@ -10,9 +10,9 @@ A high-performance Rust library for phylogenetic analysis and multiple sequence
...
+- **Maximum Likelihood Phylogenetic Analysis**: Efficient implementation of phylogenetic tree inference using SPR or NNI moves using likelihood or parsimony cost functions;
 - **Multiple Sequence Alignment (MSA)**: Support for Multiple Sequence Alignment using the IndelMaP algorithm ([paper]( https://academic.oup.com/mbe/article/41/7/msae109/7688856 ), [python implementation]( https://github.com/acg-team/indelMaP ));
-- **Sequence Evolution Models**: Support for various DNA (JC69, K80, TN93, HKY, GTR) and protein (WAG, HIVB, BLOSUM62) substitution models as well as the Poisson Indel Process (PIP) ([paper]( https://www.pnas.org/doi/10.1073/pnas.1220450110 )) model;
+- **Sequence Evolution Models**: Support for various DNA (JC69, K80, TN93, HKY, GTR) and protein (WAG, HIVB, BLOSUM62) substitution models as well as the Poisson Indel Process (PIP) ([paper]( https://www.pnas.org/doi/10.1073/pnas.1220450110 )) model and the [TKF91](https://link.springer.com/article/10.1007/BF02193625)/[TKF92](https://link.springer.com/article/10.1007/BF00163848) models;
```
junniest: Do you prefer this "convention" (paper linked to the model name) or should we stick to how I included everything else (e.g. `(PIP)([paper]( link )`)?

I think the only reason I included them separately originally is that for IndelMaP I provided the link to the Python code. We can remove that one and just link the paper for IndelMaP as you did for TKF91/92, since I would rather not have to support the Python code anyway.

README.md
```diff
@@ -72,6 +72,11 @@ This crate supports several optional features:
 - `par-regraft-manual`: Enable manual parallel regrafting control;
 - `precomputed-test-results`: Speed up test runs with precomputed results (for local development).
```
junniest: I think that, at least while it still exists, this feature should also move to `Test-related features`.

README.md
```diff
@@ -72,6 +72,11 @@ This crate supports several optional features:
...
 
+Test related features:
+
+- `recompute-brute-force-asr`: Force re-computation of brute-force ancestral sequence re-estimation in tests instead of loading precomputed results
+- `multithread-brute-force-asr`: Parallelise brute-force ancestral sequence re-estimation computation in tests
```
junniest: nitpick: punctuation, missing `;` and `.` at the end

README.md
```diff
@@ -112,11 +116,13 @@ This is a new library that is currently in active development. Contributions are
...
 
-### Current Contributors:
+### Contributors:
 
+#### Current
 - Jūlija Pečerska ([GitHub]( https://github.com/junniest ), [email]( mailto:julija.pecerska@zhaw.ch ));
 - Mattes Mrzik ([GitHub]( https://github.com/MattesMrzik ), [email]( mailto:mattes.mrzik@zhaw.ch ));
-- Dmitrii Iartsev ([GitHub]( https://github.com/jarcev ), [email]( mailto:dmitrii.iartsev@zhaw.ch ));
+#### Past
+- Dmitrii Iartsev ([GitHub]( https://github.com/jarcev ));
```
junniest, line 125: We should probably update the "citation" too. I might do that when I finally rename, but Dima left a while back by now.

I'll add an issue for the crate rename.

MattesMrzik: i updated it.

README.md
MattesMrzik: sounds good. I will remove the link to the python code and add links in the style of `[model_name](link_to_paper)`.

README.md
MattesMrzik: added a short note after the tkf models are mentioned in the `- **Sequence Evolution Models**:`
```
(the TKF models are only compatible with NNI moves, see the documentation of the `Compatible` trait);
```
Will we link the documentation somewhere? then we could also add a link directly to the Compatible trait documentation. The documentation is probably also hosted by crates.io, right?
Or do we want to host the docs of the develop branch ourself? (eg with github.io)

junniest: That's a great question. I will take a look into that today, I would like to have the docs up at least somewhere. Crates.io wants a link to the docs, but most of it is hosted on docs.rs, which needs the crate to be on crates.io... Ugh.

junniest: Made an issue for that (#161).

junniest (approved):
LGTM!

PR #162, opened 2026-03-13
Tree reading/creation fails when duplicate tip IDs are present

### LINKED ISSUE #146: Duplicate ids not detected when reading Newick or building from sequences
Duplicate leaf ids are not detected when reading Newick strings or when building a tree from sequence files.

junniest:
Closes #146.

Dealing with duplicate tip IDs when reading/creating trees.

- Creating an NJ tree will fail when the sequences have duplicate IDs;
- Newick reader will fail when tips have duplicate IDs.


phylo/src/tree/mod.rs
```diff
@@ -58,6 +58,10 @@ impl From<NodeIdx> for usize {
     }
 }
 
+pub(crate) fn generate_internal_node_id(node_idx: &usize) -> String {
+    format!("int{:?}", node_idx)
```
MattesMrzik, line 62: Would it be nice to have the `id` of the `Node` to be the same as when the node is used in the `to newick` fmt? Since here we have for examlple`int3` but when we write it to newick it will be `I3`.

phylo/src/optimisers/blen_optimiser_tests.rs
```diff
@@ -241,7 +241,7 @@ fn fix_iter() {
     let unopt_cost = c.cost();
 
     let res_default = BranchOptimiser::new(c.clone()).run().unwrap();
-    assert_eq!(res_default.iterations, 4);
```
MattesMrzik, line 244: why did this change to 3 iterations? I am just curious

phylo/src/tree/tree_parser.rs
```diff
@@ -124,6 +126,7 @@ impl Tree {
             self.add_parent_to_child_no_blen(child_idx, &Int(node_idx));
         }
         self.nodes[node_idx].children = new_children;
+        self.nodes[node_idx].id = generate_internal_node_id(&node_idx);
```
MattesMrzik, line 129: if the user named a leaf by chance like the return value of `generate_internal_node_id(&node_idx);` for some node_idx, they will get a clash. The error then will tell the user that there is a duplicate id. But if they look at their newick, they wont see that their set id clashes with the generated ones here. We should either add a comment to the `from_newick` saying how missing ids are build and that the user should avoid patterns that match `generate_internal_node_id(&node_idx);`. Or we check whether we create a duplicate and in case we get a clash then add a subscript with id to it. Sth like int3_1.

phylo/src/io/mod.rs
```diff
@@ -167,28 +168,155 @@ pub fn read_newick_from_file(path: impl AsRef<Path>) -> Result<Vec<Tree>> {
...
+    #[test]
+    fn write_newick_formatter() {
+        let trees =
+            from_newick("((A:1.0,B:2.0)X:1,(D:1.0,E:2.0)Y:1)Z:0.0;\n(A:5,B:5):10;").unwrap();
+
+        let temp_dir = tempdir().unwrap();
+        let output_path = temp_dir.path().join("output.newick");
+
+        let res = write_newick_with_formatter(&trees, output_path, |tree| tree.to_newick());
+        assert!(res.is_ok());
```
MattesMrzik, line 279: the `to_newick` is probably checked elsewhere, right? If now, then also check for the actual strings here, not only `is_ok()`.

phylo/src/phylo_info/tests.rs
```diff
@@ -63,6 +63,13 @@ fn setup_info_correct_unaligned() {
     assert!(res_info.is_ok());
 }
 
+#[test]
+fn setup_info_with_dup_ids_fails() {
+    let fldr = Path::new("./data");
+    let error = PIB::new(fldr.join("sequences_duplicated.fasta")).build();
+    assert_matches!(error, Err(Error::Sequence(msg)) if msg.contains("duplicate record id (D) found"));
```
MattesMrzik: If duplicate ids are found in trees, then we are using 'D' not (D), perhaps we should decide on one style.

phylo/src/tree/tests.rs
```diff
@@ -732,3 +806,33 @@ fn rf_distance_against_raxml() {
...
+    let error = from_newick("((A:1.0,B:1.0)X:5.1,(X:3.0,C:4.0)F:6.2)G:7.3;");
+    assert_matches!(
+        error, Err(Error::Tree(msg)) if msg.contains("node ID 'X' is not unique in the tree")
+    );
+}
+
+#[test]
+fn parse_with_wo_internal_ids() {
+    // duplicates should not be reported if internal nodes are not labeled
+    let _ = tree!("((A:1.0,B:1.0)N:5.1,(X:3.0,C:4.0):6.2):7.3;");
```
MattesMrzik: is `with_wo` intentional?

phylo/src/phylo_info/phyloinfo_builder.rs
```diff
@@ -452,21 +422,10 @@ mod private_tests {
         assert!(!ids.contains(&String::from("")));
```
MattesMrzik: i think this type of test is already covered in the tree parser tests and can be removed here.

MattesMrzik (approved):
Looks good to me. 

Perhaps it would be nice to have some plain text that explains how our newick strings must be, such that they are parsed correctly, i.e., labels start with ascii alpha and int{number} should be avoided for labels (if we decide that we dont want the subscript approach like int3_1 for clashes).

phylo/src/tree/mod.rs
junniest: Hm. The `to_newick()` will not create any labels itself, it will print what is already in the `node.id` field, so it will also print `int3` (or whatever was filled out as the ID on creation). I added a test for whether the internal ids are added upon reading newick and then doing `to_newick()` again.

phylo/src/optimisers/blen_optimiser_tests.rs
junniest: That's because I changed the parser slightly to be able to generate the ids which lead to the starting tree being slightly different and the optimisation ran faster :D the unfortunate side effect of checking changing and interdependent implementations... Not sure how to avoid that, to be honest.

phylo/src/tree/tree_parser.rs
junniest: I was contemplating that for a long time and didn't come up with a good solution for now. It's especially annoying because basically anything can be an ID so there's no guarantee we can generate something that will certainly never match a potential ID that already exists. 

My potential suggestion would be to prefix the generated ID with an underscore, i.e. `_int3`, in case there's an `int3` leaf. Though that could mess with other phylo software in case underscores are not allowed at the start of a name...

My other suggestion was to have an internal temporary identifier and a printable ID, but the NodeIdx technically already serves as an internal identifier whereas adding the IDs is pretty much only necessary for ASR output.

The other issue is that reading the sequences and the trees should be two independent steps, as these are two connected but independent structures... We could pass the known sequence IDs to the newick parser to avoid duplication, but that means that reading the tree from file is connected to the sequence data already, which I'd like to avoid.

Now in retrospect I think I prefer filling out internal node IDs as a separate step rather than during parsing or NJ tree creation, which, I think, was the way we did that before...

phylo/src/io/mod.rs
junniest: Yes, the `to_newick` is tested extensively in `tree/tests.rs`. This was just to make sure that the function `write_newick_with_formatter()` does what it's supposed to. I also moved the tests for file errors to the private test module under io, as `write_newick_with_formatter()` is private and this is the place we should check IO errors, since `write_newick_to_file` is public API and relies on this one.

phylo/src/tree/tests.rs
junniest: No, of course not, copy-paste got the best of me :D

phylo/src/phylo_info/tests.rs
junniest: Yes, good point. I fixed the message in this PR, but I will likely scratch this one and do another just for the actual bug I was trying to fix and fix this again there.

junniest: Oh dear, and some of the messages for the errors are actually "node ID 'A' is not unique in the tree".

I think this should also change to something like "duplicate node id 'A' found in the tree".

phylo/src/phylo_info/phyloinfo_builder.rs
junniest: There's actually not a designated test for that in the tree parser, so I moved it there.

phylo/src/tree/mod.rs
MattesMrzik: Okey, good. The debug print of the NodeIdx will be `I3` then, right?

junniest:
This PR has become obsolete and was superseded by PRs #177 and #179.

PR #166, opened 2026-03-25
refactor: update TKF cost builders to use parameter slices

### LINKED ISSUE #164: TKF Cost builder params as slice
Pass params not as individual params but as a slice like it's done with the PipCostBuilder. This way it's easier to handle in Jati

MattesMrzik:
## Summary
- Refactored `TKF91IndelCostBuilder`, `TKF91CostBuilder`, `TKF92IndelCostBuilder`, and `TKF92CostBuilder` to accept parameter slices (`&[f64]`) in their `new` methods, aligning them with the `PIPModel::new` pattern.
- Added parameter validation logic to handle missing values by providing defaults and logging warnings.
- Updated all test call sites and internal references to use the new slice-based API.

Closes #164

### DIFF
--- a/phylo/src/tkf_model/tests.rs
+++ b/phylo/src/tkf_model/tests.rs
@@ -320,7 +320,7 @@ pub(super) fn setup_test_phylo(alphabet: &'static Alphabet) -> PhyloInfo<MASA> {
 #[test]
 fn tkf_indel_get_and_set_params_and_freqs() {
     let mut tkf_indel_cost =
-        TKF92IndelCostBuilder::new(1.0, 2.0, 0.3, setup_test_phylo(Alphabet::dna()))
+        TKF92IndelCostBuilder::new(&[1.0, 2.0, 0.3], setup_test_phylo(Alphabet::dna()))
             .build()
             .unwrap();
     // params
@@ -348,9 +348,7 @@ fn tkf_indel_get_and_set_params_and_freqs() {
 fn tkf_get_and_set_params() {
     let subst_model = SubstModel::<GTR>::new(&[0.1, 0.2, 0.3, 0.4], &[0.5, 0.6, 0.7, 0.8, 0.9]);
     let mut tkf_cost = TKF92CostBuilder::new(
-        1.0,
-        2.0,
-        0.3,
+        &[1.0, 2.0, 0.3],
         subst_model,
         setup_test_phylo(Alphabet::dna()),
     )
@@ -388,7 +386,7 @@ fn tkf_get_and_set_params() {
 
 #[test]
 fn tkf91_indel_cost_fmt() {
-    let tkf_indel_cost = TKF91IndelCostBuilder::new(1.0, 2.0, setup_test_phylo(Alphabet::dna()))
+    let tkf_indel_cost = TKF91IndelCostBuilder::new(&[1.0, 2.0], setup_test_phylo(Alphabet::dna()))
         .build()
         .unwrap();
 
@@ -400,9 +398,10 @@ fn tkf91_indel_cost_fmt() {
 #[test]
 fn tkf91_cost_fmt() {
     let subst_model = SubstModel::<JC69>::new(&[], &[]);
-    let tkf_cost = TKF91CostBuilder::new(1.0, 2.0, subst_model, setup_test_phylo(Alphabet::dna()))
-        .build()
-        .unwrap();
+    let tkf_cost =
+        TKF91CostBuilder::new(&[1.0, 2.0], subst_model, setup_test_phylo(Alphabet::dna()))
+            .build()
+            .unwrap();
 
     let fmt = format!("{}", tkf_cost);
 
@@ -412,7 +411,7 @@ fn tkf91_cost_fmt() {
 #[test]
 fn tkf92_indel_cost_fmt() {
     let tkf_indel_cost =
-        TKF92IndelCostBuilder::new(1.0, 2.0, 0.3, setup_test_phylo(Alphabet::dna()))
+        TKF92IndelCostBuilder::new(&[1.0, 2.0, 0.3], setup_test_phylo(Alphabet::dna()))
             .build()
             .unwrap();
 
@@ -425,9 +424,7 @@ fn tkf92_indel_cost_fmt() {
 fn tkf92_cost_fmt() {
     let subst_model = SubstModel::<JC69>::new(&[], &[]);
     let tkf_cost = TKF92CostBuilder::new(
-        1.0,
-        2.0,
-        0.3,
+        &[1.0, 2.0, 0.3],
         subst_model,
         setup_test_phylo(Alphabet::dna()),
     )
@@ -443,9 +440,7 @@ fn tkf92_cost_fmt() {
 fn tkf_get_and_set_freqs() {
     let subst_model = SubstModel::<GTR>::new(&[0.1, 0.2, 0.3, 0.4], &[0.5, 0.6, 0.7, 0.8, 0.9]);
     let mut tkf_cost = TKF92CostBuilder::new(
-        1.0,
-        2.0,
-        0.3,
+        &[1.0, 2.0, 0.3],
         subst_model,
         setup_test_phylo(Alphabet::dna()),
     )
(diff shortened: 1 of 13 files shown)


phylo/src/tkf_model/tkf91.rs
```diff
@@ -170,19 +176,27 @@ impl<Q: QMatrix, AA: AncestralAlignment> TKF91CostBuilder<Q, AA> {
...
             phylo: self.phylo.clone(),
             model_info: RefCell::new(info),
         };
+
+        let subst_cost = SCB::new(self.subst_model, self.phylo.clone()).build()?;
         Ok(TKFCost {
-            indel_cost: tkf_cost,
-            subst_cost: SCB::new(self.subst_model, self.phylo).build().unwrap(),
+            indel_cost,
+            subst_cost,
```
junniest, line 190: I appreciate the removal of `unwrap()` (;

phylo/src/tkf_model/tkf91.rs
```diff
@@ -123,18 +123,26 @@ pub(super) fn validate_lambda_and_mu(lambda: f64, mu: f64) -> (f64, f64) {
...
+    pub fn new(params: &[f64], phylo: PhyloInfo<AA>) -> Self {
+        Self {
+            params: params.to_vec(),
+            phylo,
+        }
     }
 
     pub fn build(self) -> Result<TKFIndelCost<TKF91IndelModel, AA>> {
-        let (lambda, mu) = validate_lambda_and_mu(self.lambda, self.mu);
+        let mut params = self.params.clone();
```
junniest: This clone is unnecessary, since self is consumed here anyway. Another thing is that here you own (and clone) the param vector, but you still allocate two new vectors with `vec!` in lines 143 and 147.

I suggest a rewrite which only includes one allocation, and only if the `param` vector is too small:
```
        let mut params = self.params;
        if params.len() < 2 {
            warn!("Too few values provided for TKF91, 2 values required, lambda and mu");
            warn!("Falling back to default values");
            params = vec![DEFAULT_LAMBDA, DEFAULT_MU];
        } else {
            // Validate the provided parameters and adjust if necessary
            let (lambda, mu) = validate_lambda_and_mu(params[0], params[1]);
            params[0] = lambda;
            params[1] = mu;
        }
        let model = TKF91IndelModel { params };
```

The reason the validation is in the `else` branch is that if validation fails, the result is the default values, which are already set anyway. It seems pointless to validate [DEFAULT_LAMBDA, DEFAULT_MU], since if these are also not valid, then the `validate_lambda_and_mu` function shouldn't return them, but it does.

The `else` branch is still wonky, imo. The way I wrote it at least avoids another vector allocation, but it might make more sense to pass `mut params` to `validate_lambda_and_mu` and let it change them in place. 

Another tiny nitpick is that I don't think we should use `and` in function names, `validate_lambda_mu` would be enough, since the `and` just bloats the name.

junniest: And another thing is that I don't think there are tests at all for building the TKF models without given parameter values (or giving them fewer parameters than they expect). At least something basic should be there:
```
#[test]
fn tkf91_build_default() {
    let tkf_indel_cost = TKF91IndelCostBuilder::new(&[], setup_test_phylo(Alphabet::dna()))
        .build()
        .unwrap();
    assert_eq!(tkf_indel_cost.model.lambda(), DEFAULT_LAMBDA);
    assert_eq!(tkf_indel_cost.model.mu(), DEFAULT_MU);
}

#[test]
fn tkf91_build_default_one_param() {
    let tkf_indel_cost = TKF91IndelCostBuilder::new(&[0.0331], setup_test_phylo(Alphabet::dna()))
        .build()
        .unwrap();
    assert_eq!(tkf_indel_cost.model.lambda(), DEFAULT_LAMBDA);
    assert_eq!(tkf_indel_cost.model.mu(), DEFAULT_MU);
}
```

Yet another question is whether giving a single parameter value (like in `tkf91_build_default_one_param`) should still mean that both parameters are set to defaults and the value is ignored completely.

phylo/src/tkf_model/tkf91.rs
```diff
@@ -170,19 +176,27 @@ impl<Q: QMatrix, AA: AncestralAlignment> TKF91CostBuilder<Q, AA> {
             bail!(Alphabet, "alphabet mismatch between model and alignment");
         }
 
-        let (lambda, mu) = validate_lambda_and_mu(self.lambda, self.mu);
+        let mut params = self.params.clone();
```
junniest: Same comments as for `TKF91IndelModel` apply here too.

phylo/src/tkf_model/tkf92.rs
```diff
@@ -142,25 +142,27 @@ pub(super) fn validate_r(r: f64) -> f64 {
...
-            r,
+            params: params.to_vec(),
             phylo,
         }
     }
 
     pub fn build(self) -> Result<TKFIndelCost<TKF92IndelModel, AA>> {
-        let (lambda, mu) = validate_lambda_and_mu(self.lambda, self.mu);
-        let r = validate_r(self.r);
+        let mut params = self.params.clone();
```
junniest: Again, same as with `TKF91IndelModel`, feels like too many vec allocations, unnecessary clone and no tests for the "too few parameter values given" scenario. Every single test I've seen always gives the right parameter number.

phylo/src/tkf_model/tkf92.rs
```diff
@@ -206,22 +198,29 @@ impl<Q: QMatrix, AA: AncestralAlignment> TKF92CostBuilder<Q, AA> {
             bail!(Alphabet, "alphabet mismatch between model and alignment");
         }
 
-        let (lambda, mu) = validate_lambda_and_mu(self.lambda, self.mu);
-        let r = validate_r(self.r);
+        let mut params = self.params.clone();
```
junniest: Again, same comments as to `TKF92IndelCostBuilder`.

junniest (commented):
I have a few recurring comments on the TKF cost builders. I know they won't be running too frequently, but still, the number of allocations and clones seems a bit much.

phylo/src/tkf_model/tkf91.rs
MattesMrzik: good idea, changed to it

MattesMrzik: Actually, I was also able to clean up the builders that included that subst cost since previously i was not reusing the indel cost builder. Now its much better:)

phylo/src/tkf_model/tests.rs
```diff
@@ -700,27 +695,75 @@ fn tkf91_cost_builder_fails() {
...
     );
 }
 
+#[test]
+fn tkf92_build_default() {
+    let tkf_indel_cost = TKF92IndelCostBuilder::new(&[], setup_test_phylo(Alphabet::dna()))
+        .build()
+        .unwrap();
+    assert_eq!(tkf_indel_cost.model.lambda(), DEFAULT_LAMBDA);
+    assert_eq!(tkf_indel_cost.model.mu(), DEFAULT_MU);
```
junniest, line 743: Shouldn't this one also have a check for `r()`?

phylo/src/tkf_model/tkf92.rs
```diff
@@ -177,51 +194,26 @@ impl<AA: AncestralAlignment> TKF92IndelCostBuilder<AA> {
...
-            one_minus_r_over_r: (1.0 - r) / r,
-        };
-        let info = TKFIndelModelInfo::new(&model, &self.phylo);
-        let tkf_cost = TKFIndelCost {
-            model,
-            phylo: self.phylo.clone(),
-            model_info: RefCell::new(info),
-        };
+        let indel_cost = TKF92IndelCostBuilder::new(&self.params, self.phylo.clone()).build()?;
+        let subst_cost = SCB::new(self.subst_model, self.phylo).build()?;
```
junniest, line 213: This, indeed looks way better now (:

phylo/src/tkf_model/tkf92.rs
```diff
@@ -126,43 +126,60 @@ impl Display for TKF92IndelModel {
...
+        if params.len() != 3 {
+            warn!(
+                "Expected 3 parameters for TKF92 model (lambda, mu, r), but got {}",
+                params.len()
+            );
+            warn!("Falling back to default values");
+            params.resize(3, 0.0);
+            params[usize::from(TKF92Parameters::Lambda)] = DEFAULT_LAMBDA;
+            params[usize::from(TKF92Parameters::Mu)] = DEFAULT_MU;
+            params[usize::from(TKF92Parameters::R)] = DEFAULT_R;
```
junniest, line 175: Maybe it makes sense to pre-compute the indices here as well (as you did in phylo/src/tkf_model/tkf92_fixed_fragmentation.rs):
```
        let lambda_id = usize::from(TKF92Parameters::Lambda);
        let mu_id = usize::from(TKF92Parameters::Mu);
        let r_id = usize::from(TKF92Parameters::R);
```
I don't think it makes too much of a difference, but looks cleaner potentially.

junniest (approved):
LGTM!

phylo/src/tkf_model/tkf92.rs
MattesMrzik: I think i like it more this way.

PR #167, opened 2026-03-27
Fix pip freq optimization bug

MattesMrzik:
## Summary
This PR addresses a bug in the PIP frequency optimization where the initial cost was not correctly compared or maintained when empirical frequencies did not improve the log-likelihood.

- Added a regression test case `empirical_does_not_improve` in `phylo/src/optimisers/model_optimiser.rs`.
- Included supporting data (MSA and tree files) for the regression test.


MattesMrzik:
ah i wanted to create a issue not a pr

PR #168, opened 2026-03-27
Pip freq opti bug

MattesMrzik:
When freqs where set for the PIP model the `self.freqs` where set with the new params regardless of whether the subst model actually had freqs. So in the case of `JC69` and `K80` the `subst_q.q` and `subst_q.freqs` where not updated (which is correct) but the `pip_model.freqs` where udpated (incorrect). These `pip_model.freqs` are used in the logl calculation.
This pr fixes that and adds tests.

Also removed a debug assert that check if opti freqs improves logl, since with empirical freqs this is not guaranteed.


phylo/src/alphabets/mod.rs
```diff
@@ -14,7 +14,7 @@ type ConditionalProbs = DVector<f64>;
 
 pub static AMINOACIDS: &[u8] = b"ARNDCQEGHILKMFPSTWYV";
 pub static AMB_AMINOACIDS: &[u8] = b"BJZX";
-pub static NUCLEOTIDES: &[u8] = b"TCAG";
+pub static NUCLEOTIDES: &[u8] = b"TCAG"; // shouldnt we sort them alphabetically?  ACGT
```
junniest: No :D If you prefer, we can add a comment noting that this sorting is intentional to match how other tools, e.g., RaXML and PhyML, sort these. I don't think it's a good idea to switch to alphabetical (although I do wonder if changing this breaks anything...)

phylo/src/pip_model/mod.rs
```diff
@@ -120,9 +120,12 @@ impl<Q: QMatrix> EvoModel for PIPModel<Q> {
...
     fn set_freqs(&mut self, pi: FreqVector) {
         debug_assert!(self.freqs.nrows() - 1 == pi.nrows() || self.freqs.nrows() == pi.nrows());
-        self.freqs = pi.clone().insert_row(pi.nrows(), 0.0);
+        let old_subst_freqs = self.subst_q.freqs().clone();
         self.subst_q.set_freqs(pi.clone());
-        pip_q(&mut self.q, self.subst_q.q(), self.params[1]);
+        if self.subst_q.freqs() != &old_subst_freqs {
+            self.freqs = pi.clone().insert_row(pi.nrows(), 0.0);
+            pip_q(&mut self.q, self.subst_q.q(), self.params[1]);
+        }
```
junniest: I would prefer minimising unnecessary comparisons here, especially since the comparison in line 125 will only fail for two DNA models (JC69 and K80) and will just be an additional check for any other DNA model and all protein models (and for an unclear reason).

My suggestion:
```
        self.subst_q.set_freqs(pi);
        self.freqs
            .view_mut((0, 0), self.subst_q.freqs().shape())
            .copy_from(self.subst_q.freqs());
        pip_q(&mut self.q, self.subst_q.q(), self.params[1]);
```

This removes value checking and cloning; it just uses the substitution model's frequencies as the baseline. We now also avoid cloning pi and reallocating memory to extend the frequencies by modifying self.freqs in place (as we probably should have been doing from the start).

phylo/src/pip_model/tests.rs
```diff
@@ -1025,3 +1025,33 @@ fn modify_model_freqs_costs_match() {
     modify_model_freqs_costs_match_template::<BLOSUM>(new_aa_freqs.clone());
     modify_model_freqs_costs_match_template::<HIVB>(new_aa_freqs);
 }
+
+#[cfg(test)]
+fn modify_model_freqs_costs_match_template_for_model_wo_freqs<Q: QMatrix + QMatrixMaker>(
```
junniest: I have to say that these insanely long template names (and the templates themselves) are a lot. I know I started making them, but they are just an eyesore right now...

I tried to find a way to quickly shorten them to simplify, and I think I have at least one small improvement:
```

#[cfg(test)]
fn setup_test_pip_cost<Q: QMatrix + QMatrixMaker>(
    freqs: &[f64],
    params: &[f64],
) -> PIPCost<Q, MSA> {
    let info = setup_test_info(Q::alphabet());
    let model = PIPModel::<Q>::new(freqs, params);
    PIPB::new(model, info).build().unwrap()
}

#[cfg(test)]
fn modify_model_wo_freqs_costs_match_template<Q: QMatrix + QMatrixMaker>(
    freqs: FreqVector,
) {
    let mut c = setup_test_pip_cost::<Q>(&[], &[]);
    let logl = c.cost();

    c.set_freqs(freqs.clone());

    let logl2 = c.cost();
    assert_eq!(logl, logl2);

    // The likelihood should be the same if we rebuild from scratch with the same modification
    let c = setup_test_pip_cost::<Q>(freqs.as_slice(), &[]);
    let new_logl = c.cost();
    assert_eq!(new_logl, c.cost());
    assert_eq!(logl2, new_logl);
}
```

When I refactor the tests I can also use the small helper to simplify the other insane templates.

I would also suggest renaming the template func to `modify_model_wo_freqs_costs_match_template` from `modify_model_freqs_costs_match_template_for_model_wo_freqs`.

junniest (changes requested):
LGTM, but I think my proposed way to update frequencies would be an improvement, especially since it avoids cloning now.

phylo/src/alphabets/mod.rs
MattesMrzik: added the comment about RaXML and PhyML.

phylo/src/pip_model/tests.rs
MattesMrzik: implemented it. and also changed other occurrences of the pip cost creation.

phylo/src/alphabets/mod.rs
```diff
@@ -14,7 +14,7 @@ type ConditionalProbs = DVector<f64>;
 
 pub static AMINOACIDS: &[u8] = b"ARNDCQEGHILKMFPSTWYV";
 pub static AMB_AMINOACIDS: &[u8] = b"BJZX";
-pub static NUCLEOTIDES: &[u8] = b"TCAG";
+pub static NUCLEOTIDES: &[u8] = b"TCAG"; // This is legacy sorting, i.e., conforming with tools like RaXML and PhyML
```
junniest, line 17: I like this phrasing (;

phylo/src/pip_model/tests.rs
```diff
@@ -9,6 +9,8 @@ use crate::evolutionary_models::EvoModel;
 use crate::io::read_sequences;
 use crate::likelihood::ModelSearchCost;
 use crate::phylo_info::{PhyloInfo, PhyloInfoBuilder as PIB};
+#[cfg(test)]
+use crate::pip_model::PIPCost;
```
junniest: Why is this `#[cfg(test)]` necessary, and why shouldn't the use statement be merged with the next one?

junniest (approved):
LGTM, thank you!

phylo/src/pip_model/mod.rs
MattesMrzik: looks good, im using your suggestion

phylo/src/pip_model/tests.rs
MattesMrzik: its not, it was the lsp that added it. I removed the #cfg and merged the imports

PR #173, opened 2026-04-16
fix: pip_gtr_optimisation opti accuracy on ARM chip

### LINKED ISSUE #80: pip_gtr_optimisation test failing on ARM based Mac systems
# Description

When cloning from the `develop` branch and running `cargo test --features "deterministic"`, the `pip_gtr_optimisation` test fails. After `cargo clean` and recloning repository again issue persists.

## Test Error Output:
```zsh
failures:

---- optimisers::model_optimiser_tests::pip_gtr_optimisation stdout ----

thread 'optimisers::model_optimiser_tests::pip_gtr_optimisation' panicked at src/optimisers/model_optimiser_tests.rs:450:5:
assert_relative_eq!(pip_o.final_cost, -3481.828364024475, epsilon = 1e-5)

    left  = -3481.8283823856814
    right = -3481.828364024475




failures:
    optimisers::model_optimiser_tests::pip_gtr_optimisation

test result: FAILED. 339 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 378.57s
```

## System Specs:
Apple M1 Max 32GB RAM
macOS Sequoia 15.5

# Reproduction
1. Clone repository and check out the `develop` branch
2. Run `cargo test --features "deterministic"`
3. View test results

**Attached is the `debug` log level output while running this test.** (done using test-log package)

[pip_gtr_optimisation_output.txt](https://github.com/user-attachments/files/21533935/pip_gtr_optimisation_output.txt)

MattesMrzik:
Closes #80 

Floating point operations differ between chips. The ARM result of the `pip_gtr_optimisation` test was on the order of 1e-4 off from the Python script used as reference, while on the Intel chip it was only 1e-5 off.

Are the estimated parameters from the python script available? Then we could double check with the rust implementation.

### DIFF
--- a/phylo/src/optimisers/model_optimiser_tests.rs
+++ b/phylo/src/optimisers/model_optimiser_tests.rs
@@ -493,7 +493,7 @@ fn pip_gtr_optimisation() {
     assert_relative_eq!(initial_logl, -9988.840775519875, epsilon = 1e-5); // value from the python script
     assert_relative_eq!(pip_o.initial_cost, initial_logl);
     assert!(pip_o.final_cost > initial_logl);
-    assert_relative_eq!(pip_o.final_cost, -3481.828364024475, epsilon = 1e-5); // value from the python script
+    assert_relative_eq!(pip_o.final_cost, -3481.828364024475, epsilon = 1e-4); // value from the python script
     assert_eq!(pip_o.final_cost, pip_o.cost.cost());
 }
 


junniest:
If I remember correctly, I used the optimised parameter values from the model optimiser in the Rust code and just plugged them in to the Python script. The script does not actually optimise anything, it was just an example for computing PIP likelihoods.

junniest (approved):
Fair enough for now, should we keep an issue open or add a comment to the test about this difference? For now there's a bug open, but I wouldn't necessarily classify this as a bug. We could close that one but reference it in a new issue.

MattesMrzik:
I added a comment referencing #80

PR #179, opened 2026-09-07
Fix/tree from sequences duplicate ids

junniest:
This pull request refactors the construction and validation of the `Sequences` object in the alignment module. The main change is that `Sequences::new` and `Sequences::with_alphabet` now return a `Result`, and they always check for unique sequence IDs at construction time. The explicit calls to `ids_are_unique` in alignment methods have been removed, as uniqueness is now enforced earlier. Additionally, unchecked constructors are provided for test and internal use. Documentation and tests have been updated accordingly.

**API changes and validation:**

* Changed `Sequences::new` and `Sequences::with_alphabet` to return a `Result` and perform sequence ID uniqueness checks during construction, ensuring that all sequence IDs are unique when a `Sequences` object is created. (`phylo/src/alignment/sequences.rs` [[1]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R86-R96) [[2]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R111-R156)
* Added `Sequences::new_unchecked` and `Sequences::with_alphabet_unchecked` for internal/test use, which skip the uniqueness check. (`phylo/src/alignment/sequences.rs` [phylo/src/alignment/sequences.rsR111-R156](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R111-R156))

**Removal of redundant checks:**

* Removed explicit calls to `ids_are_unique` from alignment-related traits and methods, since uniqueness is now enforced by the constructor. (`phylo/src/alignment/aligner.rs` [[1]](diffhunk://#diff-d675a3600c91478eb915fa65cdb6891fe5f7cba4d3917ca8dcc2c2ae48208c6aL11-L17) `phylo/src/alignment/mod.rs` [[2]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL83-L89) [[3]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL117) [[4]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL128)

**Documentation and examples:**

* Updated documentation and doctests throughout to use the new fallible constructors and to include error handling with `Result`. (`phylo/src/alignment/sequences.rs` [[1]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R86-R96) [[2]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R111-R156) [[3]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84L134-R179) [[4]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84L147-R195) [[5]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R213-R219) [[6]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84L191-R241) [[7]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R265-R271) [[8]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R289-R294) [[9]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R307-R313) [[10]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84R347-R356)

**Test updates:**

* Updated tests to use the new fallible constructors and unchecked constructors where appropriate. (`phylo/src/alignment/sequences.rs` [[1]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84L392-L417) [[2]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84L432-R477) [[3]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84L446-R491) [[4]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84L459-R504) [[5]](diffhunk://#diff-dfcfacc7202a2480df7e8277dc75cda408bb25db6e9600a682f7b02e40cd5b84L473-R517)

**Minor adjustments:**

* Adjusted some trait implementations and doctests for consistency with the new API (e.g., using `Sequences::with_alphabet_unchecked` in `Display` for `MSA`). (`phylo/src/alignment/mod.rs` [[1]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL155-R155) [[2]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL203-R203) [[3]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL230-R230) [[4]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL258-R258) [[5]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL286-R294) [[6]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL420-R420) [[7]](diffhunk://#diff-ac4f4ce9430bfbe6cb537bef1d89373ce65fdf44fbfb1266624d365a710b758dL528-R528)


phylo/src/alignment/sequences.rs
```diff
@@ -105,14 +108,52 @@ impl Sequences {
...
+    }
+
+    /// Creates a new `Sequences` object from a vector of `bio::io::fasta::Record`.
+    ///
+    /// The alphabet is automatically detected from the sequences.
+    /// The `Sequences` object is considered aligned if all sequences have the same length.
+    /// The unchecked version is meant for test purposes only and does not check that
+    /// the ids are unique.
+    #[cfg(test)]
+    pub(crate) fn new_unchecked(s: Vec<Record>) -> Sequences {
```
MattesMrzik, line 139: Cannot we use the checked version and just call unwrap on it in tests? Perhaps sometimes when setting up toy examples some duplicate ids sneak in. Then having the check + unwrap would reveal this immediately.

phylo/src/alignment/sequences.rs
```diff
@@ -105,14 +108,52 @@ impl Sequences {
...
+        let alphabet = detect_alphabet(&s);
+        Self::with_alphabet_unchecked(s, alphabet)
+    }
+
+    /// Creates a new `Sequences` object from a vector of `bio::io::fasta::Record` and a provided alphabet.
+    ///
+    /// The `Sequences` object is considered aligned if all sequences have the same length.
+    /// The unchecked version is meant for recreating the object from an existing valid one
+    /// and does not check that the ids are unique.
+    pub(crate) fn with_alphabet_unchecked(
```
MattesMrzik, line 149: For places in the crate where we are "sure" that ids are unique it is nice to have this method. However, one could debate if avoiding the small overhead of checking it again vs catching potential bugs is better.

MattesMrzik (approved):

phylo/src/alignment/sequences.rs
junniest: That's fair, honestly I wasn't sure what's best here. I just got annoyed at having to add multiple unwraps, which often means that the code takes up an extra line. Ultimately, the tests that use this don't check that functionality, though you are right in that the examples can be buggy...

Should I remove this function then and stick to `new().unwrap()` in tests?

MattesMrzik: I guess if those tests dont rely on the ids being correct then the unchecked call can stay. For tests where the ids might be used by the methods called we still can use the checked version + unwrap call.

phylo/src/alignment/sequences.rs
junniest: The reason why I made this one `pub(crate)` is that the parsimony aligner recreates a set of sequences from the existing ones (`parsimony/mod.rs:113-124`), and those have been checked before for IDs (and should have the same alphabet). Though now I think that gapping the sequences and then using `A::from_aligned_unchecked(aligned_seqs, tree);` is kinda dumb, I would rather add a method to `Alignment` that creates an alignment from unaligned sequences and `Mapping`s, what do you think?

MattesMrzik: > The reason why I made this one pub(crate) is that the parsimony aligner recreates a set of sequences from the existing ones (parsimony/mod.rs:113-124), and those have been checked before for IDs (and should have the same alphabet)

Okay, this is to me a case where we are absolutely sure that the ids are correct. Perhaps in other cases the code might change records with their ids, then it might be worth to call the checked version. So here I agree that we use the unchecked version.

> Though now I think that gapping the sequences and then using A::from_aligned_unchecked(aligned_seqs, tree); is kinda dumb, I would rather add a method to Alignment that creates an alignment from unaligned sequences and Mappings, what do you think?

Yes, I agree that the current version is not optimal. We can create a new issue for that.

phylo/src/alignment/sequences.rs
junniest: Fair enough (as long as we remember to use it (; ) but that's also why this is a test-only function.

phylo/src/alignment/sequences.rs
junniest: Good point. I was about to try and fix it here, but you are right. I'll open an issue and merge this.

